PR #49, opened 2026-05-17
feat(controller): POST /v1/sandboxes/:id/branch — sandbox branching phase 1

WaylandYang:
## Summary

First phase of **sandbox branching** — arbitrary-depth fork-from-warm
trees. A running child sandbox can be paused, snapshot'd into a new
tag, then resumed. The resulting snapshot is independent of the
source's lifecycle: forkable from / deletable separately.

This is the controller-side surface (REST endpoint + types + tests).
CLI, Python SDK, volume inheritance, and end-to-end verification on
real KVM follow in subsequent PRs.

Design doc lives at `notes/design/branching.md` (private working
copy); moves to `docs/design/branching.md` when the feature is
user-facing in CLI/SDK.

## Use cases (agent-runtime driven)

- **Safe destructive ops** — branch before `rm -rf` / `apt remove`;
  discard on regret, keep on confirm
- **A/B tool execution** — run `grep` vs `ripgrep` in parallel
  branches, pick the winner
- **Conversation checkpointing** — auto-branch every N turns; attach
  to any past turn to reproduce a bug
- **SWE-bench-style iteration** — branch after expensive setup, iterate
  on the test prompt without re-doing `clone + install`

## Surface

```
POST /v1/sandboxes/:id/branch
  body: { "tag": "<optional>" }     # default: branch-<id>-<unix-ts>
  → 201 SnapshotInfo { tag, dir, created_at_unix, branched_from }
```

`branched_from` is a new optional field on `SnapshotInfo` carrying the
source sandbox id. Skip-serialized when `None` so existing snapshot.json
files on disk don't churn.

## Mechanism

All on top of existing primitives:

- `vm.pause()` (forkd-vmm, already present)
- `vm.snapshot_to()` (forkd-vmm, already present)
- `vm.resume()` (added in this PR; symmetric with `pause()`)

The `Vm` is taken out of `live_vms` briefly via `remove()` + reinsert
rather than holding the mutex during the slow snapshot call. Trade-off:
0.5–8 s window where `list_sandboxes` / `delete_sandbox` see the
sandbox as missing. Branching is not high-frequency; acceptable.

If a `DELETE` arrives during the branching window, the registry is the
source of truth: we drop the `Vm` on put-back if it's no longer
registered (Drop kills firecracker + cleans cgroup).

`resume()` failure after a successful snapshot is logged but does not
fail the request — the snapshot file is intact and is the user's
primary expectation. Source sandbox may end up dead in that path; this
is documented.

## Test plan

- [x] `cargo check -p forkd-controller -p forkd-vmm` clean
- [x] `cargo test -p forkd-controller --lib` — 21/21 pass (+2 new)
  - `branch_missing_sandbox_returns_404`
  - `branch_rejects_unsafe_tag`
- [ ] End-to-end on KVM testbed — defer to follow-up PR after CLI
      lands so the test scenario is realistic (`forkd branch` from a
      live child of postgres-fixture, fork grandchildren, verify
      independent psql state per grandchild)

## Pending follow-ups (separate PRs)

- CLI: `forkd snapshot --from-sandbox <id> --tag <tag>`
- Python SDK: `Sandbox.branch() → SnapshotInfo` + `Snapshot.fork(...)`
- Thread volumes from source sandbox into the branch snapshot
- `docs/API.md` — document the new endpoint + pause-window semantics
- Promote `notes/design/branching.md` → `docs/design/branching.md`
- (v0.3) Live (no-pause) branching exploration via userfaultfd

### DIFF
--- a/crates/forkd-controller/src/api.rs
+++ b/crates/forkd-controller/src/api.rs
@@ -27,6 +27,23 @@ pub struct SnapshotInfo {
     pub tag: String,
     pub dir: String,
     pub created_at_unix: u64,
+    /// Set when this snapshot was produced by branching from a running
+    /// sandbox via `POST /v1/sandboxes/:id/branch`. Carries the source
+    /// sandbox id for audit / lineage. None for snapshots built from
+    /// kernel + rootfs via `POST /v1/snapshots`.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub branched_from: Option<String>,
+}
+
+/// `POST /v1/sandboxes/:id/branch` — pause a running sandbox, snapshot
+/// it into a new tag, resume it. The resulting snapshot is independent
+/// of the source sandbox's lifecycle.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct BranchSandboxRequest {
+    /// Optional tag for the new snapshot. When unset the controller
+    /// generates `branch-<source-id>-<unix-ts>`.
+    #[serde(default)]
+    pub tag: Option<String>,
 }
 
 /// `POST /v1/sandboxes` — fork a sandbox (child VM) from a snapshot tag.
--- a/crates/forkd-controller/src/http.rs
+++ b/crates/forkd-controller/src/http.rs
@@ -14,6 +14,7 @@
 //!   POST   /v1/sandboxes/:id/ping         alive-probe through the guest agent
 //!   POST   /v1/sandboxes/:id/exec         spawn a subprocess in the sandbox
 //!   POST   /v1/sandboxes/:id/eval         eval a Python expression in PID 1
+//!   POST   /v1/sandboxes/:id/branch       pause + snapshot + resume; new tag
 //!
 //! Auth and audit logging are layered on top of this router in
 //! `lib.rs::run_daemon`. Tests in this file exercise the bare router.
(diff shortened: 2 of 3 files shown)


WaylandYang:
End-to-end verified on KVM testbed (i7-12700 / Linux 6.14 / firecracker v1.12):

```
pgfix snapshot (parent)
  └─ source sandbox (alive)
       └─ POST /v1/sandboxes/:id/branch  →  new tag, branched_from set
             ├─ POST /v1/sandboxes (n=2)  →  2 grandchildren
             │     ├─ grandchild-1: psql -c 'INSERT … SELECT count(*)' → 1 ✓
             │     └─ grandchild-2: psql -c 'INSERT … SELECT count(*)' → 1 ✓
             └─ source sandbox still listed in /v1/sandboxes after branch ✓
```

CoW isolation works across branch boundaries — each grandchild has its own postgres state independent of siblings. Source sandbox survives the pause/resume window intact.

One known limitation surfaced during the test (separate from this PR): forking N from a branch reuses netns indices `forkd-child-1..N` starting at 1, so a source sandbox occupying `forkd-child-1` collides with the first grandchild. Workaround for now is to release the source's netns before forking grandchildren; a proper fix is a separate netns-allocator change to skip in-use indices.

PR #212, opened 2026-05-31
design(v0.5): diff snapshot chains — first-pass design doc (M2.1)

WaylandYang:
## Why draft

This is a design doc for ROADMAP **M2.1 — Differential snapshots** (~3-week milestone). Posting as a draft PR for review and direction-confirmation *before* I start the implementation. Easier to redirect now than to throw away code later.

## What's in the doc

- **Motivation** — separate from v0.3's BRANCH-time diff: this is build-time chains (`python-numpy → +pandas → +sklearn`), targeting recipe-maintainer iteration time and disk usage, not BRANCH pause window.
- **Storage layout** — `snapshot.json` gains `parent_tag`; memory file is either `memory.bin` (base) or sparse `diff.bin` (chained).
- **Chain resolution** — walk parents, `cp --reflink=auto` base, merge sparse diffs in order, use topmost vmstate. Restore latency = base cp + sum-of-merges; within 10% of base on reflink-capable FS (ROADMAP done-criterion 2).
- **CLI/REST surface** — `forkd snapshot diff --from --tag --exec`, `POST /v1/snapshots/diff`, `forkd snapshot info`, `forkd snapshot compact`. `SnapshotInfo` gains `parent_tag`.
- **Hub integration** — `pack`/`unpack`/`pull` walk chains transitively. `registry.json` schema gains `parent_tag` per entry.
- **4 alternatives considered** — full re-snapshot (status quo, the thing we're replacing), FS-level CoW, OCI-style layered images (deferred to v0.6+), regenerate-rootfs-at-restore. Each with explicit rejection rationale.
- **4 open questions** — max chain depth, compact verb policy, vmstate drift across links, parent-rev pinning policy.
- **6 implementation phases** totaling ~3 weeks, aligned with ROADMAP M2.1 estimate.
- **3 risks** called out honestly: vmstate drift (largest unknown — empirical question, +1 week budget if it hits), Hub-pull disk usage on deep chains (real cost, documented not solved), `rmi`-base safety (handled via 409-with-dependents).

## What I want from review

1. **Done-criteria scope** — the 3 lifted from ROADMAP feel right; if you want different bars, easier to set them now.
2. **Storage shape** — `parent_tag` in `snapshot.json` vs. some richer manifest format. I picked simple to leave room for v0.6 CAS without locking in.
3. **The vmstate-drift risk paragraph** — if you've seen FC restore vmstate against modified RAM fail, I'd take pointers to the failure mode before Phase 2 hits it cold.
4. **Implementation phase ordering** — Phase 1 (resolver) before Phase 2 (build verb) lets me exercise the resolver against hand-crafted on-disk chains before adding the build path. Open to flipping if you'd prefer.

## Not in this PR

No code. Once direction is settled I'll land Phase 1 (snapshot.json schema + restore-side resolver) as a separate PR.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/DESIGN-v0.5-diff-snapshot-chains.md
+++ b/DESIGN-v0.5-diff-snapshot-chains.md
@@ -1,16 +1,32 @@
 # v0.5: diff snapshot chains
 
-**Status:** DRAFT — first pass. Comments via PR review or
-[`docs/ROADMAP.md` M2.1](./ROADMAP.md) tracking thread.
-**Target:** v0.5, ~3 weeks (per ROADMAP M2.1 estimate).
-**Done criteria** (lifted verbatim from ROADMAP.md):
+**Status:** DRAFT — revised after first-pass self-review (see
+[PR #212 comment](https://github.com/deeplethe/forkd/pull/212#issuecomment-4586836761)
+for what changed).
+**Target:** v0.5, 3 weeks if vmstate compat clean, 4 weeks if not
+(per ROADMAP M2.1; honest buffer included).
+
+**Done criteria** (lifted from ROADMAP.md, with explicit FS scope):
 
 1. `forkd snapshot diff --from <base-tag> --tag <new-tag>` produces a
    diff `< 100 MB` for an `apt install` / `pip install` delta.
-2. Restore time on a 3-snapshot chain is within 10% of the base
-   snapshot's restore time.
+2. Restore time on a 3-snapshot chain is within 10% of base restore
+   **on a reflink-capable filesystem** (btrfs, or ext4 with the
+   reflink feature flag, or xfs). On non-reflink FS the milestone
+   targets "within 2× of base" with `forkd doctor` warning the user.
+   See [Storage shape](#storage-shape) for why.
 3. Snapshot Hub MVP (M1.2) understands chains: pulling a diff also
    pulls its parents.
+
+**Explicitly out of done-criteria** (may still ship as auxiliary
+verbs if cheap, but not gating):
+
+- `forkd snapshot compact` — collapse a chain into a fresh base.
+  Listed in Phase 4 as an escape valve, but the milestone doesn't
+  block on it.
+- Live-fork interop. v0.5 ships file-backed chains; `live_fork:
+  true` on a chained source is documented as v0.6 work (see
+  [Live-fork interaction](#live-fork-interaction)).
 
 ## Motivation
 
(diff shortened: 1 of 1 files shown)


WaylandYang:
## Self-review pass — cross-referencing existing v0.3 code

I read this back as a reviewer. The doc isn't wrong but it materially underweights what already exists in the codebase. Here's what I'd push back on if someone else wrote it.

### 1. The BRANCH-chain pattern already ships — should be the foundation, not a footnote

[`docs/design/diff-snapshots.md`](../docs/design/diff-snapshots.md) phase 1d (\"Multi-BRANCH diff via the previous-output chain\") landed in v0.3.1. It does exactly the thing the v0.5 doc proposes — chains diff snapshots by treating each BRANCH's `memory.bin` as the next diff's base — but at BRANCH time, not build time.

`SandboxInfo.last_branch_memory_path` is the chain-head field, in `forkd-controller/src/state.rs`. `forkd-vmm::apply_diff` is the merge primitive, in `lib.rs:1553` with 3 unit tests already passing.

The design doc treats vmstate-drift as \"the biggest unknown.\" If v0.3.1 ships a chain that restores against a base + N diffs successfully (it does), then **vmstate-drift is largely settled empirically already** at least for the v0.3 chain shape. The v0.5 doc should call this out as risk-mitigated up-front, not flag it as the headline danger. The risk that *does* remain — build-time install commands creating in-kernel state not in v0.3's BRANCH path — is narrower and worth the empirical Phase 2 test, but it's not \"the whole vmstate compat question.\"

### 2. Storage model is inconsistent with the existing chain

v0.3 phase 1d stores each chained snapshot's memory as a **fully-materialized `memory.bin`** and uses the previous one as the base for the next diff in-flight. v0.5 proposes storing the chained snapshot's memory as a **sparse `diff.bin` referencing the parent**. These are two different chain shapes.

The v0.5 shape is more disk-efficient (only delta bytes on disk) but adds restore-time merge cost. The v0.3 shape is materialized (full bytes per link) so restore is just \"open one memory.bin.\"

Which shape we pick matters for done-criterion 2 (\"restore within 10% of base\"). Either pick the v0.3 materialized shape and revise the disk-savings claim, or pick the diff shape and own the restore-cost math honestly. Currently the doc tries to claim both.

**Recommendation**: pick the *diff* shape (matches user mental model of \"chain links are deltas\") but be explicit that disk savings come at a restore-time cost that requires reflink to hit done-criterion 2 on commodity HDD/SSD.

### 3. Reflink assumption is hidden in a parenthetical

The doc says \"`cp --reflink=auto` ... falls back to plain `cp` on non-CoW filesystems\" and then quietly assumes reflink for the done-criterion-2 math. On the dev box we benched on (ext4 on HDD), reflink **isn't enabled** — `cp` is a 6 s blocking copy of 1.5 GiB.

So the chain-restore latency story is:

| FS | Base restore | Chain restore (1 diff) | Delta |
|---|---|---|---|
| btrfs / xfs reflink | 1 s | ~1 s + diff merge | within 10% ✓ |
| ext4 no-reflink | 1 s | 6 s cp + diff merge | **4-6× slower** ✗ |

The doc should either: (a) explicitly require a reflink-capable FS in the recipe, with a `forkd doctor` check; (b) implement lazy materialization (mmap base + apply diffs at fault time, like v0.4 UFFD_WP does for the BRANCH path) to avoid the upfront cp altogether; (c) admit done-criterion 2 only holds on reflink-capable FS and update the criteria.

(b) is the most interesting — it's a natural reuse of the v0.4 UFFD machinery we just shipped. Probably worth a paragraph in alternatives considered.

### 4. Live-fork compatibility is unaddressed

If `python-numpy` was spawned with `live_fork: true` (memfd-backed RAM), what happens to `forkd snapshot diff --from python-numpy`? Two unanswered questions:

- Does the diff capture work the same way against a memfd-backed source as against a file-backed one?
- Can a chained snapshot be later spawned with `live_fork: true` itself?

These interact through the FC `mem_backend.backend_type: \"Shared\"` vendored patch. Either say \"yes both work, here's why\" or carve them out as v0.6 in the Non-goals section.

### 5. Phase-time-budget audit

6 phases × ~3 days each = ~18 working days = ~3.5 weeks. ROADMAP says 3 weeks. Buffer is implicit, not stated. **Recommend stating \"3 weeks if Phase 2 vmstate test is clean; 4 weeks if not\"** so we don't pretend buffer doesn't exist.

Could Phase 3 (Hub) and Phase 4 (info/compact/rmi UX) run in parallel? They touch different files. Could shave a couple of days.

### 6. Open Q4 needs a default, not a deferral

The doc says \"ships pinning by name; v0.6 enforces by hash.\" This means in v0.5, a user can break their chain by re-snapshotting a base under the same tag, silently. That's a foot-gun the doc shouldn't ship.

**Suggested default**: v0.5 records `parent_content_hash` in `snapshot.json` at build time. Restore checks the parent's current content against the recorded hash; if mismatch, fail with a clear \"chain broken — parent has been replaced; rebuild with `forkd snapshot diff --from <new-base> ...`\". Not enforce-then-fix, just enforce now.

### 7. `forkd snapshot compact` — in or out?

The doc mentions it as escape valve and lists it in Phase 4. But it isn't in the ROADMAP done-criteria. **Recommend explicitly carving it out of v0.5 done-criteria** (still ship as a verb if cheap, but don't gate the milestone on it).

### What I'd change before starting Phase 1

1. Add a section \"Prior art in v0.3\" up front, refactoring vmstate-drift risk to be informed by what already works.
2. Pick one storage shape (diff or materialized) and own the trade-offs.
3. Add reflink/lazy-materialization analysis as a third alternative (or fold into the Mechanism section).
4. Address live-fork interaction explicitly (yes or carve out).
5. Tighten Phase 1's \"5 unit tests\" → list the actual cases.
6. Commit to content-hash pinning as the v0.5 default for Open Q4.

None of these are blockers — direction is sound, just needs sharpening. Smallest useful next deliverable is still Phase 1 (resolver + schema), and the changes above can land as a doc revision before or alongside the Phase 1 code.

**My recommendation**: revise the doc once with the points above (~30 min), then start Phase 1. Or, if you'd rather see the doc-vs-code dialog play out empirically, start Phase 1 against the current doc and let the implementation push back on whichever assumptions are wrong. Either is defensible; I lean toward revise-first because the storage-shape choice (#2) is structural and changing it after Phase 1 costs rework.

WaylandYang:
Pushed self-review fixes as commit `c519943`. All 7 issues addressed:

1. Prior art section added up front; vmstate-drift risk re-scoped from "biggest unknown" to narrow install-time-state classes
2. Storage shape committed to **delta** (not materialized)
3. Reflink table makes the FS-dependence cost explicit; done-criterion 2 re-scoped to "within 10% on reflink, within 2× on non-reflink with doctor warning"
4. Live-fork interaction: build verb is file-backed only; `live_fork: true` on chained sources returns HTTP 400 in v0.5
5. New Alternative C (lazy materialization via UFFD); deferred to v0.6 with explicit reasoning
6. Phase 1: 7 concrete unit tests enumerated + 1 integration smoke
7. Content-hash pinning ships in v0.5 as the default (not deferred)

Also surfaced: phases 3+4 parallelism shaves ~2 days; honest budget is 3 weeks parallelism-clean or 4 weeks with vmstate risk.

Diff: +294 / -75. Ready for Phase 1 once direction is signed off — say the word and I'll start.

WaylandYang:
Superseded by #213, which merged the design doc together with the Phase 1 implementation. The revised design (commit `c519943`) is now on main at [`DESIGN-v0.5-diff-snapshot-chains.md`](https://github.com/deeplethe/forkd/blob/main/DESIGN-v0.5-diff-snapshot-chains.md).

PR #241, opened 2026-06-11
fix(setup): add paranoid checksum mode

### LINKED ISSUE #239: setup-host.sh: optional --paranoid mode — rustup-init binary + sha256 verify
Follow-up to #236 (thanks @jelloee for the original report — tagging you in case you want to take this one; zero pressure).

## Context

#237 documented why the curl-pipe-sh rustup install is acceptable for the default path: `rust-toolchain.toml` pins the compiler forkd actually builds with, so the rustup binary itself is low-stakes. But the comment also promised a future opt-in hardened path.

## Proposal

`scripts/setup-host.sh --paranoid`:

1. Download the `rustup-init` binary for the host triple from `https://static.rust-lang.org/rustup/archive/<version>/<triple>/rustup-init` (versioned URL, not `latest`)
2. Verify against a sha256 pinned **in the script** for the pinned rustup version
3. Run `./rustup-init -y` instead of curl-pipe-sh
4. Same treatment for the Firecracker tarball download a few lines below (it already pins `FC_VERSION` but doesn't verify a checksum — FC publishes SHASUMS per release)

Default behavior unchanged — `--paranoid` is opt-in, so the sha256-staleness tradeoff discussed in #236 only bites people who asked for it.

## Acceptance

- `setup-host.sh --paranoid` works on a clean Ubuntu 22.04/24.04 host
- Wrong sha256 → hard fail with a clear message
- README one-liner under the install section mentioning the flag

Photon101:
## Summary

Adds an opt-in `scripts/setup-host.sh --paranoid` mode for checksum-verified host setup.

## What changed

- Pins rustup-init `1.29.0` and verifies the downloaded binary for `x86_64-unknown-linux-gnu` and `aarch64-unknown-linux-gnu` before running it.
- Verifies the pinned Firecracker `v1.10.1` archive for `x86_64` and `aarch64` before extracting it.
- Preserves the default install path unless `--paranoid` is passed.
- Adds a README quick-start note showing the checksum-verified alternative.

## Verification

```bash
bash -n scripts/setup-host.sh
shellcheck scripts/setup-host.sh
bash scripts/setup-host.sh --help
bash scripts/setup-host.sh --definitely-not-real
python3 -m unittest test_analyze.py -v        # in bench/pause-window
python3 -m py_compile agent.py echo_server.py # in bench/pause-window
bash -n bench/pause-window/run.sh
cargo fmt --all -- --check
cargo clippy --all-targets --all-features -- -D warnings
cargo build --all
cargo test --all
git diff --check
```

I also exercised the paranoid download path with `HOME` set to a temp directory and `sudo` stubbed to no-op, so apt/group/KSM/hugepage operations did not mutate the host. That run downloaded and verified the Firecracker `v1.10.1` x86_64 archive, installed `firecracker` and `jailer` into the temp HOME, and confirmed `Firecracker v1.10.1`. The checksum helper was also checked against a wrong sha to confirm it hard-fails with a clear mismatch message, and against the real x86_64 rustup-init and Firecracker URLs without executing either artifact.

Closes #239

### DIFF
--- a/README.md
+++ b/README.md
@@ -460,6 +460,8 @@ curl -sSL https://github.com/deeplethe/forkd/releases/download/v0.5.2/forkd-v0.5
 
 # 2. Host bring-up:
 sudo bash scripts/setup-host.sh           # KVM + tap device, one-time
+# Or, checksum-verified host bring-up:
+# sudo bash scripts/setup-host.sh --paranoid
 sudo bash scripts/netns-setup.sh 3        # per-child network namespaces
 
 # 3. Sanity-check:
--- a/scripts/setup-host.sh
+++ b/scripts/setup-host.sh
@@ -7,6 +7,136 @@ set -euo pipefail
 say() { printf "\033[1;34m==>\033[0m %s\n" "$*"; }
 die() { printf "\033[1;31merror:\033[0m %s\n" "$*" >&2; exit 1; }
 
+usage() {
+    cat <<'EOF'
+Usage: scripts/setup-host.sh [--paranoid]
+
+Options:
+  --paranoid  Download pinned rustup-init and Firecracker archives, verify
+              their sha256 sums, then install them. Default behavior is unchanged.
+  -h, --help  Show this help text.
+EOF
+}
+
+PARANOID=0
+for arg in "$@"; do
+    case "$arg" in
+        --paranoid) PARANOID=1 ;;
+        -h|--help) usage; exit 0 ;;
+        *) usage >&2; die "unknown argument: $arg" ;;
+    esac
+done
+
+RUSTUP_VERSION="1.29.0"
+RUSTUP_X86_64_SHA256="4acc9acc76d5079515b46346a485974457b5a79893cfb01112423c89aeb5aa10"
+RUSTUP_AARCH64_SHA256="9732d6c5e2a098d3521fca8145d826ae0aaa067ef2385ead08e6feac88fa5792"
+
+FC_VERSION="v1.10.1"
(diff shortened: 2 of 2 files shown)


WaylandYang:
@Photon101 — reviewed, and this is a textbook implementation of #239. Verification I ran before merging:

1. **All 4 checksums independently reproduced** on my dev box — downloaded rustup-init 1.29.0 (both triples) and the FC v1.10.1 tarballs (both arches) fresh and hashed them myself. Exact match on all four. This was the load-bearing check for a PR like this: a wrong pin is worse than no pin.
2. **Wrong-sha hard-fail path**: fed `download_and_verify` a deliberately bad hash — clean `die` with expected-vs-got in the message, exit 1. Matches the acceptance criteria.
3. `bash -n` clean, CI green.

Things I particularly liked:

- **The `TEMP_DIRS` array + `EXIT` trap** quietly fixes a pre-existing bug — the old code leaked `$TMP` if the install died between `mktemp` and `rm -rf`. You also guarded the empty-array expansion correctly for `set -u`.
- **`host_arch()` mapping `arm64 → aarch64`** is a bonus fix: the old `uname -m` interpolation would have 404'd on some arm hosts.
- **Default path is byte-equivalent** — non-paranoid still streams `curl | tar`, so existing users see zero behavior change. Exactly the right scope discipline.
- Versioned `static.rust-lang.org/rustup/archive/` URL instead of `latest` — pin actually pins.

Merging as-is. You'll be credited in the next release notes. Welcome aboard 🚀

Issue #261, opened 2026-06-24
🐛 Potential deadlock due to blocking accept on UnixListener


asmit25805:
🐛 **Bug** · 🟠 High · Confidence: 97%

**File:** `crates/forkd-vmm/src/lib.rs`
**Location:** `request_wp_uffd`

---

### What's wrong

The line `let (stream, _) = listener.accept().context("accept FC connection")?;` blocks indefinitely if the Firecracker process never connects, causing `request_wp_uffd` to hang and the caller to block forever.

### Suggested fix

Make the listener non‑blocking and add a timeout, or use a separate thread with a bounded wait. For example:
```rust
let listener = UnixListener::bind(socket_path)
    .with_context(|| format!("bind UDS at {}", socket_path.display()))?;
listener.set_nonblocking(true).context("set non‑blocking")?;
let start = std::time::Instant::now();
let timeout = std::time::Duration::from_secs(5);
let (stream, _) = loop {
    match listener.accept() {
        Ok(pair) => break pair,
        Err(e) if e.kind() == std::io::ErrorKind::WouldBlock => {
            if start.elapsed() > timeout {
                anyhow::bail!("timeout waiting for FC to connect to UDS");
            }
            std::thread::sleep(std::time::Duration::from_millis(10));
            continue;
        }
        Err(e) => return Err(e).context("accept FC connection"),
    }
};
```

---

<details>
<summary>About this report</summary>

This finding was generated by an automated audit tool using Llama 3.3 70B + verification passes.
Only findings with ≥92% confidence that passed both LLM self-verification and line reference
verification are reported. False positives are still possible — please verify before acting.

</details>

WaylandYang:
Confirmed and fixed in #262 — thanks @asmit25805, this one is real.

`request_wp_uffd` runs `accept()` in a thread with the `UnixListener` **moved into it**, then issues the PUT that makes Firecracker dial back. If FC accepts the PUT but never connects (or the PUT fails), the blocking `accept()` never returns, so `accept_thread.join()` wedges forever — and the old "the listener drops at end of scope, so accept unblocks" comment was simply wrong (the listener is owned by the thread). Fixed with a deadline-bounded `accept_with_deadline()` (deadline = API timeout + 2 s) so a non-connecting FC surfaces as a clean error instead of a hang, plus a Linux regression test. Severity-wise it's an error-path robustness bug (the live-BRANCH path) rather than a security issue, but a real hang — good catch.

asmit25805:
[...]

Issue #260, opened 2026-06-24
🔒 Unsanitized sandbox ID can lead to path traversal in request URL


asmit25805:
🔒 **Security** · 🟠 High · Confidence: 94%

**File:** `crates/forkd-cli/src/sandbox.rs`
**Location:** `delete_sandbox`

---

### What's wrong

The sandbox ID is interpolated directly into the request URL without validation, allowing crafted IDs to target arbitrary endpoints. Problematic line: `let url = format!("{}/v1/sandboxes/{}", daemon_url.trim_end_matches('/'), id);`

### Suggested fix

Validate or encode the `id` to ensure it contains only allowed characters (e.g., alphanumeric and hyphens) before constructing the URL:
```rust
if !id.chars().all(|c| c.is_ascii_alphanumeric() || c == '-') {
    anyhow::bail!("invalid sandbox id");
}
let url = format!("{}/v1/sandboxes/{}", daemon_url.trim_end_matches('/'), id);
```

---

<details>
<summary>About this report</summary>

This finding was generated by an automated audit tool using Llama 3.3 70B + verification passes.
Only findings with ≥92% confidence that passed both LLM self-verification and line reference
verification are reported. False positives are still possible — please verify before acting.

</details>

WaylandYang:
Thanks @asmit25805 — hardened in #262, again with a precise severity note.

`delete_sandbox` is the **CLI** talking to the operator's **own** daemon. Ids from `--all`/`--tag` come from the daemon's own list response (trusted); an explicit CLI id is user-controlled, and yes, `../snapshots` would traverse the request *path* to another endpoint — but there's no privilege boundary crossed (you're hitting your own daemon, where you can already delete any sandbox you own), so this isn't really a path-traversal vuln in the security sense. The real win is a clear up-front "invalid sandbox id" error instead of a confusing 404 from a mangled path. #262 validates ids against `[A-Za-z0-9_-]` before the request, with unit tests. Appreciate the report.

asmit25805:
Thanks for the detailed explanation and the swift fix in [#262](https://github.com/deeplethe/forkd/pull/262)! Totally makes sense regarding the privilege boundary—catching the invalid ID early with a clean error is definitely a great improvement over a confusing 404. Appreciate you looking into it so quickly!

Issue #259, opened 2026-06-24
🔒 Potential directory traversal via unvalidated parent_tag


asmit25805:
🔒 **Security** · 🟠 High · Confidence: 96%

**File:** `crates/forkd-cli/src/hub.rs`
**Location:** `parent_dir`

---

### What's wrong

The code joins a potentially unsafe `parent_tag` directly onto `snap_root` without validation, allowing path traversal. The problematic line is: `let parent_dir = snap_root.join(&parent_tag);`

### Suggested fix

Validate `parent_tag` before using it in a path join, rejecting tags that contain path separators or "..". For example:
```rust
if parent_tag.contains('/') || parent_tag.contains('\\') || parent_tag == ".." || parent_tag.contains("../") {
    bail!("unsafe parent_tag '{}', possible directory traversal", parent_tag);
}
let parent_dir = snap_root.join(&parent_tag);
```

---

<details>
<summary>About this report</summary>

This finding was generated by an automated audit tool using Llama 3.3 70B + verification passes.
Only findings with ≥92% confidence that passed both LLM self-verification and line reference
verification are reported. False positives are still possible — please verify before acting.

</details>

WaylandYang:
Thanks @asmit25805 — hardened in #262, with a severity caveat worth being precise about.

The `parent_tag` here comes from a **local** `snapshot.json` that the *packer* owns, on the `forkd pack` (publishing) side. The untrusted direction — receiving someone else's pack via `forkd unpack` — already validates every chain tag with `is_safe_pack_tag` before any join (`hub.rs`), so a malicious pack can't traverse. So this isn't a cross-trust-boundary vulnerability (the "High" rating overstates it). That said, a corrupted or hand-edited local edge like `../../etc` would still escape `snap_root` on the pack-side join, so #262 applies the same `is_safe_pack_tag` check there as defense in depth. Useful flag.

asmit25805:
Thanks for the quick fix in [#262](https://github.com/deeplethe/forkd/pull/262) and for the detailed breakdown! That completely makes sense regarding the trust boundary and the unpack side already handling the validation. Appreciate you adding it as defense in depth anyway!

PR #262, opened 2026-06-24
fix: bounded WP-uffd accept (#261) + harden tag/id path joins (#259, #260)

WaylandYang:
Triage of three automated-audit reports from @asmit25805. I verified each against the actual code before acting — **one real bug + two low-severity hardenings**. The "🟠 High" ratings were overstated, but all three fixes are cheap and correct.

## #261 — real (robustness)
`request_wp_uffd` runs `accept()` in a thread with the `UnixListener` **moved into it**, then issues a PUT (10 s timeout) that makes Firecracker dial back. If FC accepts the PUT but never connects (or the PUT fails), the blocking `accept()` never returns → `accept_thread.join()` **hangs forever**. The old comment ("listener drops at end of scope → accept unblocks") was wrong — the listener is owned by the thread. Fixed with a deadline-bounded `accept_with_deadline()` (deadline = API timeout + 2 s) so a non-connecting FC is a clean error, not a hang. Linux-only regression test.

## #259 — defense in depth
`pack_chain` joined a chain `parent_tag` onto `snap_root` unvalidated. The value is from a **local** snapshot.json the packer owns (the untrusted direction — `unpack` — already validates via `is_safe_pack_tag`), but a corrupted edge like `../../etc` would escape `snap_root`. Now rejected with the same rule.

## #260 — defense in depth
`delete_sandbox` spliced a sandbox id into the request path unvalidated. Ids from `--all`/`--tag` come from the daemon's own list (trusted); an explicit CLI id is user-controlled and `../snapshots` would traverse the URL. **No privilege boundary is crossed** (CLI → operator's own daemon), but reject `[^A-Za-z0-9_-]` for a clear error instead of a confusing 404. New unit tests.

## Tests
- `accept_with_deadline` times-out test (Linux), `is_valid_sandbox_id` + `delete_sandbox`-rejects-bad-id tests.
- `cargo build/clippy -D warnings/fmt --check` clean; `cargo test -p forkd-vmm -p forkd-cli` green on the dev box.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/crates/forkd-cli/src/hub.rs
+++ b/crates/forkd-cli/src/hub.rs
@@ -325,6 +325,17 @@ pub fn pack_chain(
         if !seen.insert(parent_tag.clone()) {
             bail!("chain for `{head_tag}` reaches `{parent_tag}` twice — cycle");
         }
+        // Defense in depth (#259): parent_tag comes from a local
+        // snapshot.json the packer owns, but a corrupted/hand-edited
+        // chain edge like `../../etc` would otherwise escape snap_root on
+        // the join below. Reject anything that isn't a plain tag — the
+        // same rule the unpack side enforces on untrusted packs.
+        if !is_safe_pack_tag(&parent_tag) {
+            bail!(
+                "chain for `{head_tag}` references parent `{parent_tag}` with an unsafe tag \
+                 name (path separators / `..` not allowed) — refusing to resolve"
+            );
+        }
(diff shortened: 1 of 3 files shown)


asmit25805:
Thanks for the quick triage and the fixes, Wayland! Glad the bounded accept and the defense-in-depth path validation are in main now. Appreciate the fast turnaround!

PR #266, opened 2026-07-06
fix(security): harden daemon auth boundaries

TheusHen:
## What changed

This tightens two security edges that are easy to miss in day-to-day use:

- the controller now refuses to start unauthenticated on a non-loopback bind (`0.0.0.0`, `::`, etc.) instead of only logging a warning
- the CLI no longer prints the actual `FORKD_TOKEN` value in help output when the token is supplied via env

Both keep the local dev path boring: loopback without a token still works, and authenticated network binds still work.

## Why

These are trust-boundary footguns:

- a network-exposed daemon without auth is too dangerous to allow by default
- help text leaking bearer tokens is the kind of thing that ends up in logs, terminals, screenshots, and CI output

## Checks

- `cargo fmt --check`
- `cargo test -p forkd-cli -p forkd-controller`
- `cargo clippy -p forkd-cli -p forkd-controller -- -D warnings`
- manual: `FORKD_TOKEN=pony-secret-token-123 forkd ls --help` does not print the token value
- manual: `forkd-controller serve --bind 0.0.0.0:8899` without `--token-file` exits with `refusing unauthenticated non-loopback bind`

### DIFF
--- a/crates/forkd-cli/src/main.rs
+++ b/crates/forkd-cli/src/main.rs
@@ -107,7 +107,7 @@ enum Cmd {
         daemon_url: String,
         /// Bearer token for the controller daemon (matches `--token-file`).
         /// Read from the env var when unset.
-        #[arg(long, env = "FORKD_TOKEN")]
+        #[arg(long, env = "FORKD_TOKEN", hide_env_values = true)]
         daemon_token: Option<String>,
         /// Path to vmlinux kernel.
         #[arg(long, env = "FORKD_KERNEL", required_unless_present = "from_sandbox")]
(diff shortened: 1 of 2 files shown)


WaylandYang:
Thanks, this looks useful. I merged this into `dev` for the dev-first flow in e82270da6a761af65805d4b85bf04b576e517c86 and left `main` unchanged for now.

`cargo fmt --check` passed locally. The targeted Rust tests could not complete on my Windows box because local Cargo 1.83 resolves a transitive dependency that requires edition2024; please rely on CI / your Docker run for the Rust test signal.

Once `dev` is green, we can promote it to `main`.

TheusHen:
Thanks you for that!

Issue #263, opened 2026-06-25
Question: design rationale for tag-scoped vs sandbox-scoped branch concurrency


bouncia:
In `crates/forkd-controller/src/http.rs`, `try_acquire_branch_slot` (lines 117-139) keys the in-flight set on the *target snapshot tag* of the branch operation, not the source sandbox id:

```rust
pub fn try_acquire_branch_slot(self: &Arc<Self>, tag: &str) -> Result<BranchSlot, ...> {
    ...
    if !in_flight.insert(tag.to_string()) { return Err(AlreadyInFlight); }
    ...
}
```

The comment on `branch_in_flight` (lines 53-58) explains why: "two concurrent `POST /branch` calls targeting the same tag from racing to clobber memory.bin". That's clearly correct — two clients picking the same explicit `tag` shouldn't be allowed through.

But the *default* tag is auto-generated as `branch-<sandbox-id>-<unix-now-secs>` (line 531). So the realistic collision case is two clients branching the same sandbox at the same second — which gets rejected with 409 "branch for tag 'branch-…' is already in progress". From the client's POV that's surprising: they didn't pick the same tag, the daemon did, and the daemon then refused them.

Two design questions:

1. **Would per-sandbox-id exclusion be more useful?** A given source sandbox can only have one in-flight pause/snapshot/resume cycle anyway (pause is exclusive — line 599 takes the `Vm` out of `live_vms`), so keying on sandbox id rather than tag gives the same correctness with a clearer error ("sandbox X is already being branched"), and the auto-tag collision goes away.

2. **Should the auto-tag include sub-second precision?** Even with tag-scoped locking, generating tags that include `unix_now()` in seconds means two branches of the same sandbox in the same second collide. A nanosecond suffix or a short random would prevent that without changing the semantics.

I have my own preference (sandbox-id exclusion + sub-second auto-tag) but don't want to send a PR without knowing the maintainers' rationale for the current shape. Happy to write either version.

WaylandYang:
Thanks for the careful writeup — verified all of it against current `main`. Your reasoning is sound; the line numbers drifted (`http.rs` moved a lot since you filed). Re-anchoring, then a disposition.

[...]

Issue #257, opened 2026-06-23
Question: why MAP_PRIVATE for first BRANCH instead of userfaultfd for everything?


Bug-Keeper:
Reading the design — `MAP_PRIVATE` mmap of `memory.bin` as the CoW primitive, then optionally `userfaultfd` (v0.3 scaffolding per `docs/design/userfaultfd.md`) for live branching. The current code lands on `MemoryBackend::File` as default and `MemoryBackend::Userfault` is bail-only (`crates/forkd-vmm/src/lib.rs:1035-1040`).

Two related questions for the design discussion:

1. **What's holding you back from defaulting Userfault for spawn, not just BRANCH?** The prewarm flag (`prewarm_scratch_dir`) is explicitly there to compensate for the cold-cache penalty of MAP_PRIVATE — which userfaultfd dodges entirely because pages are served on demand. If the handler is implemented, would there be any reason to keep File as the default at all?

2. **Why a per-child uffd handler socket** rather than a daemon-wide one? The `Userfault { handler_sock: PathBuf }` shape in `MemoryBackend` (`lib.rs:284-289`) implies one socket per child. A single handler with a routing layer could amortize the metadata bookkeeping across all children of a tag — particularly relevant for the 100-children fanout numbers in the README. I'd expect the design doc to address this trade-off but couldn't find it.

I'm asking because I'm thinking about contributing to the uffd path for v0.3 and want to make sure I don't go in the wrong direction.

WaylandYang:
Good questions — and the short answer probably saves you a wrong turn: the `MemoryBackend::Userfault` arm you're reading is v0.3 scaffolding that was never wired into the restore path. It's referenced in exactly two places — the enum variant and a `bail!` arm in `restore_many_with` — so setting it just errors out. Don't build on it.

The live-fork primitive that actually shipped is the `mode="live"` BRANCH path (`forkd-uffd`'s `WpBranch`): it arms `UFFDIO_WRITEPROTECT` on the running source's guest memory and asynchronously copies dirty pages. It's built on top of the `MemoryBackend::MemfdShared` backing — a per-child MAP_SHARED memfd (`"shared": true`) served by the vendored Firecracker fork, which the controller also `mmap`s so it can observe the guest's writes. That's the code to contribute to — note `MemfdShared` is the memory backing, the primitive is `WpBranch`.

[...]

PR #267, opened 2026-07-08
Add KSM doctor check and memfd probe evidence

### LINKED ISSUE #5: KSM directed hints for fork families
## Problem
Linux's default KSM scan is too lazy — minutes to reach steady-state sharing. We need seconds.

Dirty problem #6 in DESIGN.md.

## Approach
- `madvise(MADV_MERGEABLE)` on memory.bin VMA immediately after each restore.
- Tune `pages_to_scan` and `sleep_millisecs` based on family size.
- Stretch: kernel patch for fork-aware KSM (declare known-shared instead of scanning).

## Acceptance
- [ ] Within 1 s of restore, `Shared_Clean / Rss > 90%` for the memory.bin region
- [ ] Doesn't break existing KSM accounting

TheusHen:
## Summary

This updates the #5 follow-up after checking the KSM counters directly.

The original version of this PR tried to mark a controller-side `MAP_SHARED` memfd mapping as `MADV_MERGEABLE`. After the review, I added a small probe under `bench/ksm-memfd/` and ran it against `/sys/kernel/mm/ksm` counters. The result matches the concern from review: the anonymous private control case merges, but the memfd `MAP_SHARED` case does not move `pages_shared` or `pages_sharing`.

So this PR now removes the misleading memfd-side hint instead of keeping a no-op mapping around. What remains is:

- a `forkd doctor` KSM row, so operators can see when KSM is disabled or has `pages_to_scan=0`;
- a reproducible KSM probe plus captured results showing that memfd `MAP_SHARED` is not the right implementation path for #5;
- small docs cleanup around #257 and #263.

This does not claim to fully fix #5 anymore. It narrows the next implementation attempt to a mapping/backing shape KSM actually scans.

## Issues

Helps #5 by adding the doctor check and ruling out the memfd `MAP_SHARED` approach.
Documents #257 and #263.

## Testing

- `cargo fmt --check`
- `git diff --check`
- `cargo test --workspace`
- `cargo clippy --workspace --all-targets -- -D warnings`
- `python -m py_compile sdk/mcp/forkd_mcp/server.py bench/ksm-memfd/probe.py`
- `docker run --rm --privileged -v "$PWD:/work" -w /work python:3.12-slim bash -lc 'python bench/ksm-memfd/probe.py --pages 4096 --timeout 30'`

### DIFF
--- a/bench/ksm-memfd/RESULTS.md
+++ b/bench/ksm-memfd/RESULTS.md
@@ -1,0 +1,51 @@
+# KSM memfd probe
+
+This probe checks whether `MADV_MERGEABLE` produces KSM sharing for
+memfd-backed `MAP_SHARED` regions.
+
+Why this exists: issue #5 asks for directed KSM hints for fork families.
+For the v0.4 live-fork path the tempting implementation is to mark the
+controller's memfd mapping mergeable. The Linux KSM documentation says
+KSM targets private anonymous pages, though, so this benchmark keeps the
+evidence next to the code before forkd grows a misleading hint.
+
+## Environment
+
+- Run from Docker with `--privileged`
+- Host KSM sysfs writable at `/sys/kernel/mm/ksm`
+- Probe command:
+
+```bash
+docker run --rm --privileged \
+  -v "$PWD:/work" -w /work \
+  python:3.12-slim \
+  bash -lc 'python bench/ksm-memfd/probe.py --pages 4096 --timeout 30'
+```
+
+The probe saves and restores `run`, `pages_to_scan`, and
+`sleep_millisecs`. Before each case it writes `run=2` to unmerge any
+existing KSM pages, waits for `pages_sharing=0`, then enables KSM and
+waits for scans.
+
+## Result
+
+```text
+anonymous-private
+  before={'pages_shared': 0, 'pages_sharing': 0, 'pages_unshared': 0, 'pages_volatile': 0, 'full_scans': 0}
+  after ={'pages_shared': 2, 'pages_sharing': 510, 'pages_unshared': 0, 'pages_volatile': 5069, 'full_scans': 2}
+  delta ={'pages_shared': 2, 'pages_sharing': 510, 'pages_unshared': 0, 'pages_volatile': 5069, 'full_scans': 2}
+memfd-map-shared-two-fds
+  before={'pages_shared': 0, 'pages_sharing': 0, 'pages_unshared': 0, 'pages_volatile': 0, 'full_scans': 0}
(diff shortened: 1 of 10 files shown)


WaylandYang (changes requested):
Thanks for tackling #5. I think this needs one more proof point before merge.

The implementation marks a controller-side `MAP_SHARED` memfd mapping as `MADV_MERGEABLE`, but Linux KSM documentation says KSM only merges private anonymous pages and does not merge page-cache/file-backed pages. Since memfd/shmem `MAP_SHARED` pages appear to fall on the wrong side of that boundary, this may be a no-op apart from adding a populated controller mapping.

Could you add evidence from `/sys/kernel/mm/ksm/pages_shared`, `pages_sharing`, etc. showing that memfd-backed fork families actually merge with this change?

TheusHen:
No problems.

I will verify and add right now.

TheusHen:
Thanks, you were right to ask for the counters here.

I added `bench/ksm-memfd/probe.py` and ran it with KSM enabled from a privileged Docker container. The anonymous private control case does merge:

```text
anonymous-private
  delta ={'pages_shared': 2, 'pages_sharing': 510, 'pages_unshared': 0, 'pages_volatile': 5069, 'full_scans': 2}
```

But the memfd `MAP_SHARED` case does not move the KSM counters at all:

```text
memfd-map-shared-two-fds
  delta ={'pages_shared': 0, 'pages_sharing': 0, 'pages_unshared': 0, 'pages_volatile': 0, 'full_scans': 0}
```

So I removed the controller-side memfd `MADV_MERGEABLE` implementation rather than keeping a no-op populated mapping. The PR now keeps the useful `forkd doctor` KSM check and adds the probe/results as evidence that this memfd path is not the right #5 implementation.

WaylandYang (approved):
Thanks for checking the counters and changing direction. This version is much better: it removes the no-op runtime memfd hint, keeps the useful doctor signal, and leaves a reproducible probe/results trail for why the MAP_SHARED memfd path is not the right #5 implementation.

I checked the updated diff: no remaining runtime MADV_MERGEABLE path in forkd-vmm/controller, `cargo fmt --check` passes locally, and `bench/ksm-memfd/probe.py` passes `py_compile`.

WaylandYang:
Merged this updated version into `dev` as ed4b7f8fcf5f06b46304639050787492529d9562, keeping #266 already on `dev` as well.

Local checks:
- `cargo fmt --check` passed.
- Rust tests are still blocked on my Windows box by local Cargo 1.83 resolving a dependency that requires edition2024.
- Python syntax check could not run here because `python` is not on PATH in this shell; the probe is Linux/Docker-oriented and your reported Docker run is the relevant signal.

Leaving `main` unchanged until dev is promoted.

PR #280, opened 2026-08-08
fix(cli): include size + extras in rootfs cache key

jrimmer:
**Related issue: #284**

---

The rootfs cache key was derived only from the image name (slug), so `forkd from-image elixir:1.17` (default 1536 MiB) and `forkd from-image elixir:1.17 --size-mib 3072` both mapped to the same `elixir-1-17.ext4`. A rebuild with a different `--size-mib` (or `--extra`) silently reused the stale rootfs, producing a truncated/corrupt ext4 that kernel-panics on mount (`VFS: Unable to mount root fs`).

`rootfs_cache_key()` hashes `size_mib` + `extra` packages into the cache filename → `elixir-1-17-3072-<sha8>.ext4`. Applied in both `from_image_cmd` and `run_cmd`. Adds 4 unit tests (size, extras, determinism, slug sanitization).

Verified: `cargo test -p forkd-cli` → 36 passed; re-baked `elixir:1.17 --size-mib 3072` after the fix → snapshot builds and boots, `elixir --version` → 1.17.3 in a live sandbox.

### DIFF
--- a/crates/forkd-cli/src/main.rs
+++ b/crates/forkd-cli/src/main.rs
@@ -1815,6 +1815,35 @@ fn images_cmd() -> Result<()> {
     Ok(())
 }
 
+/// Build a deterministic cache filename for a rootfs from the image name,
+/// size, and extra packages. Including size + extras in the key means a
+/// rebuild with different flags does not silently reuse a stale rootfs
+/// (e.g. a too-small default build cached under the same image slug).
+fn rootfs_cache_key(image: &str, size_mib: u32, extra: &[String]) -> String {
+    use sha2::{Digest, Sha256};
+    let slug: String = image
+        .chars()
+        .map(|c| if c.is_alphanumeric() { c } else { '-' })
+        .collect::<String>()
+        .trim_matches('-')
+        .to_string();
+    let mut hasher = Sha256::new();
+    // Hash the RAW image reference with a length prefix so two distinct
+    // refs that slug to the same filename (e.g. "foo/bar:1" and
+    // "foo-bar-1") cannot collide: the slug is only for readability, the
+    // digest is the collision-proof part of the key.
+    hasher.update((image.len() as u64).to_le_bytes());
+    hasher.update(image.as_bytes());
+    hasher.update(size_mib.to_le_bytes());
+    for pkg in extra {
+        hasher.update(pkg.as_bytes());
+        hasher.update([0]);
+    }
+    let digest = hasher.finalize();
+    let short: String = digest.iter().take(8).map(|b| format!("{b:02x}")).collect();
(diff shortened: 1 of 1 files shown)


jrimmer:

jrimmer:

3. **Formatting.** Running the guideline's `cargo fmt --all -- --check` locally flagged two formatting deviations in this branch's changed lines; `cargo fmt --all` was applied and the amendment includes the formatted result.

WaylandYang (changes requested):
Thanks for the fix and the added tests. One cache-correctness blocker remains: `rootfs_cache_key` feeds only `size_mib` and `extra` into SHA-256; the image reference is represented only by the lossy slug. Distinct valid image references such as `foo/bar:1` and `foo-bar-1` both normalize to `foo-bar-1`, so identical size/extras will still reuse the same rootfs and can boot content from the wrong source image.

Please include the raw image reference in the hashed payload (with unambiguous framing, e.g. length prefixes or separators) and add a regression test covering two image names that collide after slug sanitization.

jrimmer:
Thanks for the review — confirmed, and fixed in commit `733ffa9`.

The digest previously covered only `size_mib` + extra packages; the image reference appeared *only* in the lossy slug filename, so two distinct refs that slug identically (e.g. `foo/bar:1` vs `foo-bar-1`) with the same size/extras produced the **same cache key** and could boot the wrong rootfs.

**Fix:** hash the raw image reference into the digest, length-framed (`len(image) || image`), before the existing size/extras. The slug stays in the filename purely for human readability; the digest is the collision-proof part of the key.

**Regression test:** `cache_key_differentiates_refs_with_same_slug` asserts `foo/bar:1` and `foo-bar-1` (same size/extras) produce different keys, and that the same ref stays deterministic.

**Tradeoffs:** the length prefix means two refs that are byte-different but share a prefix can't alias (e.g. `ab` + `c` vs `a` + `bc`); SHA-256 truncation to 8 bytes for the filename keeps keys short — collision odds there are negligible for a cache namespace, and full-digest verification still applies to the rootfs content when it's mounted.

fmt/clippy/tests green.

jrimmer:
History rebuilt: the branch is now a single clean commit on top of current main (5e457c4) — no merged-upstream history in the diff. Same content as before: raw image ref hashed (length-framed) into the rootfs cache key + slug-collision regression test. Force-pushed; the old commits are gone from the branch.

WaylandYang (approved):
Re-reviewed the cleaned-up head. The cache key now covers the raw image reference, size, and NUL-delimited extras; the slug-collision regression is correct, the diff is scoped again, and CI is green. Approved.

PR #281, opened 2026-08-09
fix(controller): serialize concurrent spawns to avoid tap collision

jrimmer:
**Related issue: #285**

---

## Problem

When two `POST /v1/sandboxes` spawns run concurrently, both firecracker processes race to open the single shared host tap (`forkd-tap0`). One fails with:

```
Open tap device failed: Resource busy (os error 16)
```

The existing retry loop in `create_sandbox` then re-attempts against a firecracker that was **already started**, surfacing:

```
The requested operation is not supported after starting the microVM.
```

That corrupt sandbox is unusable — any subsequent exec against it fails with `Connection reset by peer`. In a CI runner that grants sandboxes concurrently (e.g. multiple workflow jobs), this intermittently fails jobs.

## Fix

Hold a spawn mutex across the restore in `create_sandbox`, so concurrent spawns serialize and never collide on the tap. Each spawn still restores its N children in parallel internally, so this is a correctness fix, not a throughput cap.

Uses `tokio::sync::Mutex` (not `parking_lot`) so the guard is `Send` and the axum handler future stays `Send` for the `Handler` bound.

## Testing

- `cargo test -p forkd-controller`: 78 passed, 0 failed
- Full workspace: only pre-existing `forkd-vmm` chain test fails (unrelated, fails on clean main too)

### DIFF
--- a/crates/forkd-controller/src/http.rs
+++ b/crates/forkd-controller/src/http.rs
@@ -61,6 +61,28 @@ pub struct AppState {
     /// of memory.bin; without a cap, an attacker can fill the disk by
     /// firing many BRANCHes in parallel.
     pub branch_sem: Arc<Semaphore>,
+    /// Owner lease for the single shared host tap (`forkd-tap0`),
+    /// enforced by the controller (security review #281).
+    ///
+    /// `forkd-tap0` is a host tap that can be attached by ONE live
+    /// firecracker at a time. A mutex only serialized the restore call —
+    /// it did NOT prevent a second VM from grabbing the tap after the
+    /// first VM was already live (the first VM still holds the tap fd
+    /// for its whole lifetime, so the second spawn would hit "Resource
+    /// busy" even after acquiring the mutex). This lease makes that
+    /// invariant explicit: while a shared-tap VM is live, a second
+    /// shared-tap request is rejected with 503. `per_child_netns=true`
+    /// spawns do NOT touch this (each child netns has its own tap), so
+    /// parallel per-netns sandboxes are unaffected.
+    ///
+    /// `parking_lot::Mutex<Option<String>>` where the value is the owning
+    /// sandbox id. The claim is made INSIDE the `spawn_blocking` task
+    /// (so a cancelled async handler cannot release it mid-restore) and
+    /// committed in the async handler AFTER `live_vms.insert()` (so a
+    /// cancelled handler between restore success and registration does
+    /// not wedge the tap permanently). Released when the VM leaves
+    /// `live_vms` (delete / suspend).
+    pub shared_tap_owner: std::sync::Arc<parking_lot::Mutex<Option<String>>>,
     /// The configured maximum the `branch_sem` was constructed with.
     /// Tracked separately for `/metrics` (`forkd_branch_concurrency_cap`)
     /// because `Semaphore` doesn't expose its initial permit count.
@@ -1135,6 +1157,32 @@ async fn create_sandbox(
     let spawn_live_fork = req.live_fork;
     let spawn_hugepages = req.hugepages;
     let tag_for_work_dir = tag.clone();
+    // Shared-tap lease (review #281): for per_child_netns=false spawns,
+    // claim the shared host tap INSIDE spawn_blocking so handler
+    // cancellation cannot release it mid-restore. The claim is returned
+    // UNCOMMITTED and committed AFTER live_vms.insert() so a cancelled
+    // handler between restore success and registration does not wedge
+    // the tap permanently. The token is the first child's sandbox id;
+    // release_shared_tap_if_owner checks against it on teardown.
+    let use_shared_tap = !req.per_child_netns;
+    let tap_owner_cell = s.shared_tap_owner.clone();
+    let tap_token = if use_shared_tap {
+        Some(new_sandbox_id())
+    } else {
+        None
+    };
+    // Reject shared-TAP batches above n=1 (review #281 round 3): when
+    // per_child_netns=false, all children share a single host tap fd
+    // (forkd-tap0). The tap lease is owned by the first child's sandbox
+    // id; deleting that child releases the lease while sibling VMs
+    // remain live on the same tap fd, causing EBUSY on the next
+    // shared-tap spawn. The common case is n=1; n>1 shared-TAP spawns
+    // are rejected with 503 until per-child tap ownership is modeled.
+    if use_shared_tap && req.n > 1 {
+        return service_unavailable(
+            "shared-TAP spawns with n>1 are not supported; use per_child_netns=true for multi-child spawns",
+        );
+    }
 
     // v0.5 chain resolution. Resolve the chain BEFORE spawn_blocking
     // because it only reads snapshot.json files (cheap). The expensive
@@ -1175,7 +1223,8 @@ async fn create_sandbox(
     let prewarm_requested = req.prewarm;
     let mut snapshot = snapshot;
     let head_tag_for_log = req.snapshot_tag.clone();
-    let (fork_result, mut netns_reservation) = match tokio::task::spawn_blocking(move || -> anyhow::Result<(forkd_vmm::ForkResult, Option<crate::netns::NetnsReservation>)> {
+    let tap_token_for_closure = tap_token.clone();
+    let (fork_result, mut netns_reservation, mut tap_claim) = match tokio::task::spawn_blocking(move || -> anyhow::Result<(forkd_vmm::ForkResult, Option<crate::netns::NetnsReservation>, Option<SharedTapClaim>)> {
         // Reserve the netns offset range INSIDE the blocking task so
         // the reservation survives async-handler cancellation. If
         // the handler is dropped while this task runs, the task still
@@ -1188,6 +1237,23 @@ async fn create_sandbox(
             match netns_alloc.reserve(spawn_n) {
                 Some(r) => Some(r),
                 None => return Err(anyhow::Error::new(NetnsExhausted)),
+            }
+        } else {
+            None
+        };
+        // Claim the shared tap (if this spawn uses it) before restore.
+        // A second live shared-tap VM is rejected with SharedTapBusy ->
+        // 503; the claim is returned UNCOMMITTED and committed by the
+        // async handler after live_vms.insert() so handler cancellation
+        // between restore success and registration does not wedge the
+        // tap permanently (review #281).
+        let tap_claim = if use_shared_tap {
+            match try_claim_shared_tap(
+                &tap_owner_cell,
+                tap_token_for_closure.expect("use_shared_tap => token"),
+            ) {
+                Some(c) => Some(c),
+                None => return Err(anyhow::Error::new(SharedTapBusy)),
             }
         } else {
             None
@@ -1272,7 +1338,7 @@ async fn create_sandbox(
                 std::thread::sleep(std::time::Duration::from_millis(backoffs_ms[attempt - 1]));
             }
             match snapshot.restore_many_with(opts.clone(), &work_dir) {
-                Ok(r) => return Ok((r, netns_reservation)),
+                Ok(r) => return Ok((r, netns_reservation, tap_claim)),
                 Err(e) => {
                     let msg = format!("{e:#}");
                     let is_busy = msg.contains("Resource busy")
@@ -1280,14 +1346,8 @@ async fn create_sandbox(
                         || msg.contains("os error 16");
                     if !is_busy || attempt == backoffs_ms.len() {
                         return Err(e);
+
                     }
-                    tracing::warn!(
-                        attempt = attempt + 1,
-                        next_backoff_ms = backoffs_ms[attempt],
-                        error = %e,
-                        "restore_many: tap/cgroup busy, retrying"
-                    );
-                    last_err = Some(e);
                 }
             }
         }
@@ -1295,8 +1355,11 @@ async fn create_sandbox(
     })
     .await
     {
-        Ok(Ok((fr, nr))) => (fr, nr),
+        Ok(Ok((fr, nr, tc))) => (fr, nr, tc),
         Ok(Err(e)) if e.downcast_ref::<NetnsExhausted>().is_some() => {
+            return service_unavailable(&format!("{e:#}"));
+        }
+        Ok(Err(e)) if e.downcast_ref::<SharedTapBusy>().is_some() => {
             return service_unavailable(&format!("{e:#}"));
         }
         Ok(Err(e)) => return server_error(&format!("restore_many: {e:#}")),
@@ -1315,10 +1378,11 @@ async fn create_sandbox(
 
     let now = unix_now();
     let mut infos = Vec::with_capacity(fork_result.children.len());
+    let mut first_id_override = tap_token.clone();
     {
         let mut live = s.live_vms.lock();
         for vm in fork_result.children {
-            let id = new_sandbox_id();
+            let id = first_id_override.take().unwrap_or_else(new_sandbox_id);
             let info = SandboxInfo {
                 id: id.clone(),
                 snapshot_tag: tag.clone(),
@@ -1347,6 +1411,13 @@ async fn create_sandbox(
         if let Some(res) = netns_reservation.as_mut() {
             res.commit();
         }
+        // Commit the shared-tap lease AFTER live_vms.insert() (review #281):
+        // committing inside spawn_blocking would wedge the tap permanently
+        // if the handler is cancelled between restore success and
+        // registration.
+        if let Some(claim) = tap_claim.as_mut() {
+            claim.commit();
+        }
     }
 
     (StatusCode::CREATED, Json(infos)).into_response()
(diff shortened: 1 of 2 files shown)


jrimmer:

jrimmer:

WaylandYang (changes requested):
The mutex does not make allocation/spawn atomic, so the reported collision is still reachable:

1. `netns_offset` and `work_dir` are computed before acquiring `spawn_mutex`. Two concurrent per-child-netns requests can both choose the same offset/work directory, then run serially with the second using its stale choice after the first VM has been registered.
2. Workspace create/resume goes through `spawn_one_for_workspace` and bypasses this mutex entirely.
3. For the shared-TAP path, serializing only the restore attempt does not release the first live VM's TAP fd; the next active VM can still hit `Resource busy` after it acquires the mutex.

Please centralize spawn allocation as an atomic reservation/lease shared by sandbox and workspace paths, hold it through successful `live_vms` registration (with rollback on failure), and define the shared-TAP policy so a second live owner cannot be launched. Add a concurrent-spawn regression test that would fail with the current pre-lock allocation.

jrimmer:
Thanks — and apologies for the earlier commit attribution noise; the branch has been rewritten so all commits are authored by me (jrimmer).

You're right on both points, and they're fixed across commits `7ec4a89` + `aacc4cc` (stacked on #282):

1. **`netns_offset` is now computed inside the critical section.** In `create_sandbox`, `spawn_mutex` is acquired *before* the netns reservation, so the offset pick, the Firecracker restore, and the `live_vms` registration are one atomic unit — a concurrent per-child-netns request can no longer see a stale offset. (The allocator from #282 also makes the pick itself atomic; the mutex additionally guarantees a unique `work_dir` per batch.)
2. **The workspace path is serialized too.** `create_workspace` and `resume_workspace` both call `spawn_one_for_workspace`, which restores onto the same shared `forkd-tap0` — but neither took `spawn_mutex`, so the tap race survived via the workspace path. Both now take the same mutex across spawn + registration.

**Shared-TAP policy (explicit):** a second live owner cannot be launched. The single host tap is opened during restore; `spawn_mutex` guarantees at most one restore in flight across *all* spawn paths (sandbox, workspace create, workspace resume), so no two VMs race for `forkd-tap0`. This is the "global serialize" option — the conservative choice that keeps one shared tap.

**Tradeoff / future opportunity (option b):** if per-VM TAPs become desirable (parallel spawn throughput without serialization), the allocator pattern from #282 (atomic reservation + RAII lease, injectable pool) is exactly the shape a per-VM tap allocator would take — `NetnsAllocator` generalizes to `TapAllocator` with the same lease semantics, and `spawn_mutex` could then be dropped in favor of per-tap leases. Happy to pursue that as a follow-up if upstream prefers parallel spawns over the global serialize; the current change keeps the invariant (no tap collision) with minimal surface area.

fmt/clippy/tests green.

jrimmer:
History rebuilt: the branch is now a single clean commit on top of current main (5e457c4). It is no longer stacked on #282 — this PR is standalone (spawn_mutex serialization + workspace-path coverage + shared-TAP policy), and #282 carries the atomic netns allocator. Content unchanged; force-pushed.

WaylandYang:
Thanks for rebuilding the branch; the diff is scoped again and CI is green. The shared-TAP ownership blocker remains, though.

spawn_mutex serializes restore calls, but it does not enforce shared-TAP ownership for the lifetime of a live VM. After the first restore and registration complete, the guard is released while that Firecracker process can still own forkd-tap0; a second request may then acquire the mutex and still fail with EBUSY. Thus the statement that a second live owner cannot be launched is not enforced by this implementation.

There is also a cancellation gap: the mutex guard remains in the async handler rather than being owned by the spawn_blocking task. If the handler is cancelled, the guard is dropped while the blocking restore continues, allowing another restore to overlap it.

For per_child_netns=true, distinct namespaces contain distinct same-named TAPs, so globally serializing those requests is unnecessary and can hide the allocator problem addressed by #282.

Please first land a corrected #282 and re-check the original reproduction. If a shared-TAP path still needs protection, either maintain an explicit owner lease until VM teardown and reject or queue another shared-TAP owner, or allocate distinct TAP/netns ownership. Please add a regression test with the first shared-TAP VM still live when the second request arrives, plus coverage for cancellation/guard lifetime; a test covering only simultaneous handler entry is insufficient.

jrimmer:
Thanks for the thorough review — all three points verified, and the fix is now in (commit `b8d089f`).

**You were right: a mutex that only serializes the restore call does NOT enforce tap ownership.** The first VM keeps `forkd-tap0` open for its whole lifetime, so a second spawn could acquire the mutex and still hit `Resource busy`. The guard also lived in the async handler, so a cancelled request dropped it while `spawn_blocking` was still restoring.

**What changed — explicit owner lease:**

1. **Owner lease held until VM teardown.** `SharedState` now carries `shared_tap_owner: Arc<Mutex<Option<String>>>` (the live owner's sandbox id). A second shared-tap spawn is **rejected with 503** ("shared host tap forkd-tap0 is in use by another live sandbox; delete it first or use per_child_netns=true for parallel sandboxes") instead of racing into an EBUSY retry loop. Release happens only when the VM leaves `live_vms`: `delete_sandbox`, `delete_workspace`, `suspend_workspace`.
2. **Cancellation-safe claim.** The claim is made INSIDE the `spawn_blocking` task (a `std::sync::Mutex` cell held by the blocking closure), committed only after restore succeeds, and released on drop if the spawn fails. A cancelled async handler cannot release the claim while the blocking restore continues — that closes the guard-lifetime gap you identified.
3. **per_child_netns=true skips the lease entirely.** Each child netns has its own tap, so globally serializing those requests was unnecessary; parallel per-netns sandboxes are unaffected (regression test asserts a per-netns spawn is NOT rejected while the shared tap is owned).
4. **Shared by all spawn paths.** `create_sandbox`, `create_workspace`, and `resume_workspace` all go through the same claim — the workspace path previously bypassed the mutex entirely, which is what let a workspace resume race `create_sandbox`.

**Tests added:** second shared-tap owner while the first is live → 503 (names `forkd-tap0` and preserves the original owner); per-netns ignores the lease; lease lifetime semantics (claim blocks a second claimer, uncommitted drop releases, commit keeps the lease until explicit teardown). `cargo fmt`, `clippy -D warnings`, and the full controller suite are green.

**On the shared-tap policy itself:** the lease enforces "a second live owner cannot be launched" by rejecting rather than queueing — explicit, immediate, and honest about the one-owner-per-tap kernel constraint. The alternative (allocate distinct TAP per sandbox, i.e. per-VM taps) is a clean follow-up and the better long-term answer for multi-sandbox hosts; it's a bigger change (tap lifecycle + netns wiring) so I'd suggest tracking it separately. Happy to open that as an issue if you'd like.

jrimmer:
### Cross-PR merge conflict (resolved)

#281 retained `pick_netns_offset` which #282 deletes. Both modified the same `http.rs` lines and `spawn_one_for_workspace`'s return type. **This branch is now rebased on `fix/bound-netns-allocator` (#282)** — `pick_netns_offset` is gone, and `spawn_one_for_workspace` now returns `(Vm, SandboxInfo, Option<NetnsReservation>, Option<SharedTapClaim>)` with both leases uncommitted. The async caller commits both after `live_vms.insert()`.

### P1: Tap lease committed inside spawn_blocking before registration (fixed)

The tap lease was committed inside the `spawn_blocking` closure immediately after restore succeeded — before `live_vms.insert()`. If the handler was cancelled (client disconnect) between restore success and registration, the committed lease was never released, permanently locking the tap (DoS until restart).

**Fix:** The `SharedTapClaim` is now returned UNCOMMITTED from the blocking task and committed by the async handler AFTER `live_vms.insert()` — mirroring #282's netns reservation pattern. This preserves cancellation-safety during restore (drop releases if restore fails) and closes the post-restore leak.

### P1: Teardown releases tap lease before killing VM (fixed)

`delete_sandbox`, `delete_workspace`, and `suspend_workspace` released the tap lease before `drop(vm)` killed firecracker — a new spawn could claim the tap and hit EBUSY while the old process was still dying.

**Fix:** All teardown paths now `drop(vm)` first (kill firecracker), then release both the shared-tap lease and netns index. Uses `release_shared_tap_if_owner(s, &sandbox_id, is_shared_tap)` which checks the owner token matches the departing VM's sandbox ID.

### P2: `shared_tap_owner` used `std::sync::Mutex` (poisoning risk) (fixed)

A panic while holding it would poison the mutex and panic all subsequent spawns.

**Fix:** Switched to `parking_lot::Mutex` (matching `netns.rs` and the rest of `AppState`).

### P2: `release_shared_tap_if_owner` cleared unconditionally (no token check) (fixed)

The teardown release did `if vm.netns.is_none() { *owner = None }` — never compared the cell's token to the departing VM's ID. The token was also the snapshot tag, not the sandbox ID.

**Fix:** `release_shared_tap_if_owner` now takes `(s, &sandbox_id, is_shared_tap)` and only clears when `owner.as_deref() == Some(sandbox_id)`. The token is now the sandbox ID (was the snapshot tag), which is pre-generated before the tap claim so it matches what release checks.

### P2: Per-netns offset race (resolved by rebase)

#281 still used `pick_netns_offset` (check-then-act) for per-netns spawns. Rebasing on #282 replaces it with the atomic `NetnsAllocator`, closing this race.

### Tests

- `create_sandbox_rejects_second_shared_tap_owner_with_503` — second shared-tap spawn returns 503, lease stays with original owner
- `create_sandbox_per_netns_ignores_shared_tap_lease` — per-netns spawn bypasses the lease
- `shared_tap_lease_claim_release_and_drop_semantics` — claim/drop/commit lifecycle (uses `parking_lot::Mutex`)

WaylandYang (changes requested):
The explicit owner cell fixes the cross-request restore race, but the current shared-TAP lifetime is still incorrect for batches and take-out operations:

1. For `n > 1 && per_child_netns=false`, one request starts multiple Firecracker children against the same `forkd-tap0` while holding one owner token. This does not prevent sibling collisions inside `restore_many_with`. If the batch does start, the token is assigned to the first child ID; deleting that first child clears the lease even though its siblings remain live on the shared tap. Either reject shared-TAP batches above one or model ownership for every live child/use distinct TAPs.
2. BRANCH temporarily removes the VM from `live_vms` without transferring the shared-TAP lease into an RAII owner. Cancellation, or DELETE removing the registry entry during that window, drops the VM after the task but never clears `shared_tap_owner`, wedging all future shared-TAP spawns until restart. Workspace suspend has the same cancellation dependency.
3. This branch inherits #282''s analogous netns cancellation/leak paths, so it should remain stacked and blocked until those ownership semantics are corrected.

Please add tests for a shared-TAP `n=2` batch and deletion of each sibling, plus cancellation and DELETE during BRANCH/SUSPEND.

WaylandYang (commented):
One lifecycle gap remains across controller process restarts. On startup, `Registry::reconcile()` keeps records whose PIDs are alive, but `live_vms` and `shared_tap_owner` are initialized empty. If the controller crashes while Firecracker survives, the restarted controller therefore considers `forkd-tap0` unowned and can admit a new shared-TAP VM even though the orphan still owns it. The retained registry record is not enough: DELETE has no reconstructed `Vm` handle with which to terminate that process, and a bare `/proc/<pid>` existence check is vulnerable to PID reuse.

Please define and test restart recovery: either verify process identity, terminate orphan Firecracker processes, and remove their registry entries before accepting spawns; or genuinely adopt them and rebuild TAP ownership/live handles from durable state. Until then, the lease is only correct within one controller process lifetime.

jrimmer:
## Round 3 fixes — N>1 shared-TAP rejection and BRANCH/suspend tap lease cancellation

All CHANGES_REQUESTED findings from the 2026-08-12 review are addressed in commit e55ca5b (rebased onto #282 r3).

### 1. Reject shared-TAP batches with n>1 (503)
When `per_child_netns=false`, all children share a single host tap fd. The tap lease is owned by the first child's sandbox id; deleting that child releases the lease while sibling VMs remain live, causing EBUSY. The common case is n=1; n>1 shared-TAP spawns are now rejected with 503 until per-child tap ownership is modeled.

### 2. Extend `VmNetnsGuard` to release shared-tap lease on Drop
The guard now carries optional `shared_tap_owner` + `tap_owner_id`. On Drop (cancellation or DELETE during BRANCH/suspend), it clears the owner if it matches. This ensures the tap lease is always released when the VM is killed, surviving handler cancellation. Both `branch_sandbox` and `suspend_workspace` pass the shared-tap info to the guard.

### 3. Rebase onto #282 r3
The shared-tap lease (SharedTapClaim, try_claim_shared_tap, release_shared_tap_if_owner) is rebased on top of #282's cancellation-safety changes. The tap claim is made inside `spawn_blocking` alongside the netns reservation, returned uncommitted as part of a 3-tuple `(ForkResult, Option<NetnsReservation>, Option<SharedTapClaim>)`, and committed after `live_vms.insert`. Both `NetnsExhausted` and `SharedTapBusy` are mapped to 503.

### Controller restart recovery gap (COMMENTED finding)
Tracked as a follow-up issue: #298.

jrimmer:
## CI fix

Rebased on #282's latest tip (includes the `netns_reservation` clippy fix). Two issues resolved:

1. **Inherited unused `mut`** on `netns_reservation` — fixed by rebase onto #282.
2. **Dead `last_err` variable** — during the original rebase, the `last_err = Some(e)` assignment was lost from the retry loop's busy-retry branch. Since the loop always returns on every iteration (Ok on success, Err on non-busy or final attempt), `last_err` was dead code and the post-loop `Err(last_err.expect(...))` was unreachable. Removed the variable entirely and replaced with `unreachable!()`.

All round 2 CHANGES_REQUESTED findings remain addressed:
- N>1 shared-TAP spawns rejected with 503 (use `per_child_netns=true` for parallel spawns)
- `VmNetnsGuard` extended with shared-tap lease release for BRANCH/suspend cancellation safety
- `SharedTapClaim` committed after `live_vms.insert` (not inside `spawn_blocking`)
- `parking_lot::Mutex` for `shared_tap_owner` (poison-free)
- `release_shared_tap_if_owner` checks owner token before clearing

The COMMENTED controller-restart gap is tracked separately in #299.

CI is now green across all 5 checks. Requesting re-review.

WaylandYang (changes requested):
One ownership-transfer bug remains in VmNetnsGuard::into_vm(). The successful BRANCH path takes the Vm and then mem::forget(self), so every reinsertion into live_vms permanently leaks the guard fields: the NetnsAllocator Arc, shared_tap_owner Arc, and tap_owner_id String. Repeated BRANCH operations make this an unbounded daemon-lifetime leak.

Please let the guard be dropped normally after taking the VM, and make Drop release netns/TAP ownership only when self.vm is still Some (meaning the guard still owns and kills the VM). This preserves the active leases on successful transfer while dropping the bookkeeping allocations. Please add a regression test that observes the Arc strong counts or an equivalent drop sentinel across repeated into_vm transfers. This branch should remain stacked on the corrected #282 and merge after it.

jrimmer:
## Fix: VmNetnsGuard::into_vm() Arc leak

The old `into_vm()` called `std::mem::forget(self)` after taking the VM, permanently leaking the guard's `Arc<NetnsAllocator> and the shared-tap Arc` on every successful BRANCH reinsertion. A long-running controller would accumulate one leaked allocation per BRANCH without bound.

**Fix:**
- `into_vm()` now takes the VM out (sets `self.vm = None`) and lets the guard drop normally instead of `mem::forget`. The remaining fields (`Arc<NetnsAllocator> and the shared-tap Arc` and bookkeeping strings) are released on drop, decrementing refcounts — no leak.
- `Drop` checks `self.vm.is_some()` before killing the VM and releasing the netns index. When `into_vm()` was called, `self.vm` is `None`, so `Drop` skips the kill+release path (the index must stay ACTIVE because the VM was transferred to `live_vms`).

**Regression test:** `vmnetnsguard_into_vm_does_not_leak_arc` constructs 50 consumed guards (the post-`into_vm()` state) and drops them, then asserts `Arc::strong_count` returns to baseline. If the old `mem::forget` behavior were still present, the count would stay elevated.

WaylandYang (changes requested):
Thank you for resolving the shared-TAP ownership and Arc-lifetime findings. #282 has now landed on main, but GitHub cannot rebase this branch automatically because the stacked allocator history conflicts with the merged implementation. Please rebase onto current main and retain only #281's shared-TAP lease/batch-validation changes; do not reintroduce the old allocator commits. Once the resulting diff is scoped, CI is green, and the ownership/cancellation tests still pass, I expect this to be ready for final approval.

jrimmer:
Rebased onto current `main` now that the netns allocator (#282) has landed.

The branch now contains only the shared-TAP lease and batch-validation changes:

- the old allocator commits are gone (they are on `main` via #282),
- the `VmNetnsGuard::into_vm` Arc-leak fix is also on `main` via #282, so this branch no longer carries it,
- shared-TAP `n>1` rejection, the `SharedTapBusy` 503 path, the owner-token check in `release_shared_tap_if_owner`, and the `VmNetnsGuard` shared-tap release on drop are all retained.

Verified: `cargo clippy` clean and the controller test suite passes (89 tests).

WaylandYang (changes requested):
Thank you for rebasing this onto current main and removing the duplicated allocator history. I re-reviewed the resulting diff: the shared-TAP owner lease, `n > 1` shared-TAP rejection, owner-token check, and cancellation/teardown transfer through `VmNetnsGuard` now look correct, and CI is green. I do not see a remaining code-level blocker in this scoped PR.

One repository requirement still blocks approval: commit `768de13` has no `Signed-off-by` trailer. README's contribution instructions require every commit to be signed off with `git commit -s`. Please rewrite/squash the branch so every resulting commit carries your DCO sign-off, force-push, and let CI rerun. After that, this should be ready for approval.

WaylandYang (approved):
Re-reviewed the current head. The shared-TAP owner lease is held through spawn and registration, n > 1 shared-TAP use is rejected, cancellation safely releases reservations, and delete/teardown ordering prevents early reuse. CI is green and I found no remaining correctness blocker. Approved.

PR #283, opened 2026-08-10
fix(controller): persist per_child_netns across workspace resume

jrimmer:
**Related issue: #287**

---

## Problem

`resume_workspace` hardcodes `per_child_netns: false`, respawning the live sandbox on the shared `forkd-tap0` even when the workspace was created with a per-child netns (`per_child_netns: true`).

With several active workspaces this causes:
- **Tap collisions** — a second restart can fail with `Open tap device failed: Resource busy … forkd-tap0`
- **Broken proxy routing** — downstream consumers that `setns` into the sandbox's netns get `setns: invalid argument` / 502s because the endpoint has no netns path

## Fix

- Add `per_child_netns: bool` to `WorkspaceInfo` with `#[serde(default)]` — existing state files stay backward compatible
- Set it at `create_workspace` from the request field
- Honor it at `resume_workspace` instead of hardcoding `false`

## Notes

- No behavior change for existing workspaces (they default to `false`, matching current resume behavior)
- `spawn_one_for_workspace` already takes the flag; only the resume call site ignored it

### DIFF
--- a/crates/forkd-controller/src/api.rs
+++ b/crates/forkd-controller/src/api.rs
@@ -400,6 +400,12 @@ pub struct WorkspaceInfo {
     /// `suspend?diff=true`.
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub last_branch_memory_path: Option<std::path::PathBuf>,
+    /// Whether this workspace's live sandboxes use a per-child netns.
+    /// Persisted so that `resume` respawns on the same netns layout
+    /// the workspace was created with (the shared forkd-tap0 is not a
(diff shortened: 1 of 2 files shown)


jrimmer:

jrimmer:

WaylandYang (approved):
Reviewed the persisted network-layout plumbing and backward-compatible serde default. The change is scoped to the reported resume regression, and the maintainer-approved CI run passes fmt, clippy, build, and the full test matrix. Existing pre-change workspaces defaulting to shared networking is explicitly documented; they can be recreated to opt in.

PR #289, opened 2026-08-12
fix(vmm): force clocksource=kvm-clock to prevent stale TSC after snapshot restore

### LINKED ISSUE #288: Sandbox VMs have stale system clock after snapshot restore (breaks TLS/SSL)
## Problem

Sandbox VMs created via snapshot restore have a stale system clock. The clock reflects the snapshot creation time rather than the current host time, causing it to drift by days or weeks depending on snapshot age.

This was discovered when setting up automated LLM-based PR review in [spoond](https://github.com/jrimmer/spoond). The review workflow runs inside a forkd sandbox VM and makes HTTPS calls to an external LLM API. The SSL handshake fails with:

```
certificate is not yet valid
```

Investigation showed the sandbox VM's system clock was ~75 hours (3+ days) behind the host. The SSL certificate had been issued the same day, so the VM's stale clock fell before the certificate's `notBefore` date.

## Root Cause

Firecracker VM kernels initially use `kvm-clock` (host-synced) but switch to `tsc` early in init (visible in `dmesg`: `clocksource: Switched to clocksource tsc`). The TSC value in a forkd snapshot reflects the snapshot creation time, not the current host time. Since forkd creates sandboxes by restoring from snapshots, the VM clock starts at the snapshot time and advances from there. Clock staleness is proportional to snapshot age.

The shipped 4.14 `vmlinux` kernel exhibits this behavior: kvm-clock is available and used initially, but TSC overrides it during kernel init.

### Option A: Force `clocksource=kvm-clock` in kernel boot args (this PR)

Append `clocksource=kvm-clock` to the kernel boot args in both `BootConfig` constructors (`quickstart()` and `ext4_rw()`). This prevents the kernel from switching to TSC, so the guest always reads the host's wall clock via KVM paravirtualized clock MSRs.

**Pros:**
- Addresses the root cause (prevents TSC selection)
- Clean, declarative fix in boot configuration
- No runtime overhead

**Drawbacks:**
- **Forward-only on snapshots:** Boot args are applied only during the original VM boot that creates a snapshot. Firecracker's snapshot restore path (`PUT /snapshot/load`) does not re-apply `PUT /boot-source`, so existing snapshots created before this change retain the stale-TSC kernel state. **All existing snapshots must be re-baked via `forkd from-image`** before restored sandboxes inherit the fix.
- x86_64-only: `kvm-clock` (`CONFIG_KVM_CLOCK`) is an x86 paravirtualized clock. On aarch64 the parameter is silently ignored (no compile-time or runtime guard currently).

### Option B: Set guest clock after snapshot restore

After `PUT /snapshot/load` completes and the VM resumes, exec `date -s <host_time>` inside the VM via the guest agent (forkd-agent.py on port 8888). The guest agent is already running in the snapshot state and resumes along with the VM.

**Pros:**
- Works for **all** existing snapshots immediately — no re-bake required
- Not forward-only: takes effect on every restore
- Directly fixes the symptom (wrong clock) regardless of clocksource

**Drawbacks:**
- **Band-aid, not root cause fix:** The clock is wrong for the first few seconds after restore until the guest agent responds and `date -s` completes. Any process that starts during this window sees a stale clock.
- Adds a network round-trip (guest agent ping + exec) to every sandbox creation
- Best-effort: if the guest agent is unresponsive, clock sync silently fails

### Recommendation

The PR (#289) implements Option A (the kernel boot-args fix) as the root-cause fix. Option B could be added as a complementary defense-in-depth measure for existing snapshots that haven't been re-baked yet, but is not included in this PR.

jrimmer:
## Summary

Related to #288. This is a defense-in-depth fix for modern x86 hosts; it does not fully close #288 on older kernels or aarch64.

Forces `clocksource=kvm-clock` in both guest kernel boot configurations to prevent the kernel from switching to a TSC that is stale after snapshot restore.

## Scope

- Adds `clocksource=kvm-clock` to `quickstart()` and `ext4_rw()`.
- Keeps it as the final `clocksource=` override.
- Adds unit coverage plus an ignored boot → snapshot → wait 20s → restore hardware test with a 2s drift tolerance.
- Detects whether the host kernel supports `KVM_CLOCK_REALTIME` (Linux 5.16+).

This is forward-only for snapshots: existing snapshots must be re-baked with `forkd from-image`. Older kernels and aarch64 still require a general restore-time correction, so #288 must remain open.

## Verification

- Rust CI passes on the reviewed implementation.
- The real Firecracker test remains `#[ignore]` because it requires Linux, KVM, Firecracker, a kernel image, rootfs, and TAP setup.

### DIFF
--- a/crates/forkd-vmm/src/lib.rs
+++ b/crates/forkd-vmm/src/lib.rs
@@ -2512,4 +2512,216 @@ mod tests {
         // WorkDirGuard drops work_dir.
         drop(vm);
     }
+
+    /// Integration test: verify kvm-clock survives snapshot/restore on
+    /// modern x86 kernels (≥ 5.16 with KVM_CLOCK_REALTIME).
+    ///
+    /// This test complements boot_vm_uses_kvm_clock (which tests fresh
+    /// boot only). The snapshot-restore path is where stale TSC actually
+    /// manifests: after restore, KVM_SET_CLOCK restores the saved clock
+    /// state from the snapshot. On Linux ≥ 5.16, the KVM_CLOCK_REALTIME
+    /// flag makes kvm-clock auto-correct to host wall time, so the boot
+    /// arg alone is sufficient. On older kernels (< 5.16) or aarch64,
+    /// the restored clock is stale — see PR #300 (guest-side date -s)
+    /// for the complementary fix.
+    ///
+    /// This test verifies #289's contribution: the boot arg keeps
+    /// kvm-clock as the clocksource through the snapshot/restore cycle,
+    /// and on modern kernels the restored clock is correct without any
+    /// guest-side intervention.
+    #[test]
+    #[ignore = "requires firecracker, kernel, rootfs, and tap device"]
+    #[cfg(target_os = "linux")]
+    fn kvm_clock_survives_snapshot_restore() {
+        let kernel = std::env::var("FORKD_TEST_KERNEL")
+            .unwrap_or_else(|_| "/var/lib/forkd/kernels/vmlinux".to_string());
+        let rootfs = std::env::var("FORKD_TEST_ROOTFS")
+            .expect("FORKD_TEST_ROOTFS must point to an ext4 rootfs image");
+
+        assert!(
+            std::path::Path::new(&kernel).exists(),
+            "FORKD_TEST_KERNEL not found at {kernel}"
+        );
+        assert!(
+            std::path::Path::new(&rootfs).exists(),
+            "FORKD_TEST_ROOTFS not found at {rootfs}"
+        );
+
+        let boot_dir =
+            std::env::temp_dir().join(format!("forkd-restore-test-boot-{}", std::process::id()));
+        let snap_dir =
+            std::env::temp_dir().join(format!("forkd-restore-test-snap-{}", std::process::id()));
+        let restore_dir =
+            std::env::temp_dir().join(format!("forkd-restore-test-restore-{}", std::process::id()));
+
+        let _ = std::fs::remove_dir_all(&boot_dir);
+        let _ = std::fs::remove_dir_all(&snap_dir);
+        let _ = std::fs::remove_dir_all(&restore_dir);
+
+        let _boot_guard = WorkDirGuard(boot_dir.clone());
+        let _snap_guard = WorkDirGuard(snap_dir.clone());
+        let _restore_guard = WorkDirGuard(restore_dir.clone());
+
+        let addr = "10.42.0.2:8888";
+
+        assert!(
+            ping_at(addr).is_err(),
+            "{addr} is already reachable — a stale VM may occupy the address; \
+             kill it before running this test"
+        );
+
+        // --- Phase 1: Boot fresh VM and verify kvm-clock ---
+
+        let cfg = BootConfig::ext4_rw(kernel.into(), rootfs.into(), boot_dir.clone())
+            .with_network(NetworkConfig::default_tap("forkd-tap0"));
+
+        let vm = Vm::boot(&cfg).expect("boot VM");
+
+        let mut connected = false;
+        for _ in 0..60 {
+            std::thread::sleep(std::time::Duration::from_millis(500));
+            if ping_at(addr).is_ok() {
+                connected = true;
+                break;
+            }
+        }
+        assert!(
+            connected,
+            "guest agent at {addr} did not respond within 30s; \
+             check console log at {}/fc.console",
+            boot_dir.display()
+        );
+
+        // Verify clocksource on fresh boot
+        let resp = exec_at(
+            addr,
+            vec![
+                "cat".into(),
+                "/sys/devices/system/clocksource/clocksource0/current_clocksource".into(),
+            ],
+            std::time::Duration::from_secs(10),
+        )
+        .expect("exec clocksource check");
+        assert_eq!(resp.exit_code, 0, "cat clocksource failed: {}", resp.stderr);
+        assert_eq!(
+            resp.stdout.trim(),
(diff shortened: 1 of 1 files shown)


WaylandYang (approved):
Reviewed the current head. The change is scoped to both boot configurations, keeps kvm-clock as the final clocksource override, documents the forward-only snapshot behavior, and CI is green. The ignored hardware test does not yet exercise a full snapshot/restore cycle, which would be useful follow-up coverage, but I do not see a correctness blocker in this patch.

WaylandYang (changes requested):
Blocking: the integration test only verifies a fresh boot and never exercises the failing snapshot-restore path.

forkd vendors Firecracker v1.12, whose snapshot documentation explicitly states that the guest wall clock continues from snapshot creation. Its x86 restore path restores the saved KVM clock with KVM_SET_CLOCK; selecting kvm-clock does not inherently read the host wall clock.

On Linux >= 5.16, KVM_CLOCK_REALTIME may make the vendored v1.12 behavior appear to work, but forkd also supports older 5.10/5.15 hosts where that flag is unavailable. Upstream Firecracker only made this behavior explicit through the clock_realtime restore option in firecracker-microvm/firecracker#5809.

Please add a real boot -> snapshot -> wait -> restore regression test and implement a solution that covers the documented host-kernel range, such as guest-side clock correction or an explicit vendored clock-realtime restore mechanism. The boot argument alone is not a complete fix.

jrimmer:
## Update: snapshot-restore integration test added

Added `kvm_clock_survives_snapshot_restore` `#[ignore]` integration test to address the CHANGES_REQUESTED finding:

1. Boots a fresh VM with `clocksource=kvm-clock` boot arg
2. Verifies clocksource=kvm-clock on fresh boot
3. Pauses and snapshots the VM
4. Kills the source VM (frees tap device + guest agent address)
5. Waits 3 seconds (any stale clock would be visibly wrong)
6. Restores from the snapshot via `restore_many_with`
7. Verifies clocksource is still kvm-clock after restore
8. Verifies wall clock matches host time (within 5s tolerance)

On modern x86 (≥5.16), `KVM_CLOCK_REALTIME` makes kvm-clock auto-correct to host wall time on restore, so the boot arg alone is sufficient — this test verifies that. On older kernels (<5.16) or aarch64, the restored clock would be stale; that gap is addressed by PR #300 (guest-side `date -s` after restore), which has its own complementary test using a deliberately-wrong mock clock to isolate the scripting solution.

### Relationship between #289 and #300

The two PRs fix different aspects of the same problem:

- **#289** (this PR): Forces `clocksource=kvm-clock` at boot. On modern x86, `KVM_CLOCK_REALTIME` auto-corrects on restore. Forward-only — existing snapshots must be re-baked.
- **#300**: Runs `date -s` in each restored guest via the guest agent. Works on all kernels/architectures and existing snapshots. Best-effort with a brief wrong-clock window.

They are complementary — #289 handles modern x86 cleanly (no guest-side intervention), #300 handles everything else. The recommended deployment is both: #289 for the common case, #300 as a fallback.

WaylandYang (changes requested):
The new restore test still cannot distinguish a working KVM_CLOCK_REALTIME correction from no correction. It waits only 3 seconds after snapshot creation but accepts up to 5 seconds of host/guest drift, so a guest that simply resumes at the saved timestamp can pass.

Please make the wait comfortably larger than the allowed error (for example, wait 15-30 seconds and require <=2 seconds drift), and detect the host kernel/capability so the test has explicit behavior on the documented 5.10/5.15 range. The implementation remains a modern-x86 optimization rather than a complete fix for #288: forkd documents support for host kernels where KVM_CLOCK_REALTIME is unavailable. Either narrow the PR/issue-closing claim to defense-in-depth for modern x86, or pair it with a proven general restore-time correction before treating #288 as fixed.

jrimmer:
## Fix: restore clock test hardened, #288 claim narrowed

The restore test waited 3 seconds but allowed up to 5 seconds of drift — a guest that simply resumed at the saved timestamp could pass by accident, making the test unable to distinguish a working KVM_CLOCK_REALTIME correction from no correction.

**Changes:**
- Increased the wait before restore from 3s to 20s (comfortably larger than the allowed 2s drift), so a stale clock cannot pass by accident.
- Tightened the allowed drift from 5s to 2s.
- Added `host_kernel_release()` helper that reads `/proc/sys/kernel/osrelease` to detect the host kernel version.
- Added `kernel_supports_kvm_clock_realtime(release)` helper that parses the release string and returns true for kernel ≥ 5.16 (where KVM_CLOCK_REALTIME makes kvm-clock auto-correct on restore).
- The test now branches on the detected kernel: on modern kernels (≥ 5.16), the drift assertion is a hard requirement; on older kernels (5.10/5.15), the assertion still fires but the message points to PR #300 for the complementary guest-side fix, making the limitation explicit rather than silent.
- Added a unit test for the kernel version parser covering 5.10/5.15/5.16/6.1 and edge cases.

**Scope clarification:** The boot-arg fix (clocksource=kvm-clock) is defense-in-depth for modern x86. On kernels < 5.16 where KVM_CLOCK_REALTIME is unavailable, the restored kvm-clock stays stale and the guest-side `date -s` correction (PR #300) is required. #288 is not fully resolved until #300 also lands — this PR addresses the modern x86 path and pairs with #300 for older kernels.

WaylandYang (approved):
Thanks for tightening the restore test and narrowing the scope. The 20-second wait with a 2-second tolerance now detects the original false positive, and the PR no longer claims to close #288 outside modern x86/KVM_CLOCK_REALTIME hosts. Approved pending the rerun CI; #288 should remain open for the general restore-time fix.

PR #292, opened 2026-08-12
fix(agent): use container PATH from /etc/environment for subprocess calls

### LINKED ISSUE #290: forkd-agent.py inherits host PATH instead of container environment
## Problem

`forkd-agent.py` (PID 1 inside every sandbox VM) executes commands via the `exec` action using `subprocess.run(args, capture_output=True, timeout=timeout)` with **no `env=` parameter**. The subprocess inherits whatever PATH the agent process itself has — which comes from the host environment that launched firecracker, not from the container's `/etc/environment`.

This means container-installed tools at non-standard locations are invisible:

- **Go**: `golang:1.25` Docker images install `go` to `/usr/local/go/bin/` — not in the inherited PATH → `go: command not found` (exit 127)
- **Rust**: `rust:1.85-bookworm` installs `cargo` to `/usr/local/cargo/bin/` — same issue
- **Any tool** a Docker image installs outside the default `/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin` is affected

## Impact

This caused CI-breaking failures in spoond's Forgejo Actions CI: after a fresh Docker rebuild of the `go-base` forkd snapshot, CI test steps failed with `go: command not found` because the agent's inherited PATH didn't include `/usr/local/go/bin`.

## Proposed fix

Parse `/etc/environment` at agent startup to get the container's PATH. Build a `GUEST_ENV` dict with the container PATH as default, and pass `env=_subprocess_env(cmd.get("env"))` to all `subprocess.run`/`Popen` calls. Caller-provided env vars merge on top of `GUEST_ENV` so they can still override PATH if needed.

Specifically:
1. Add `_load_container_env()` that parses `/etc/environment` (pam_env format: `KEY=VALUE` lines, no `export` keyword)
2. Build `GUEST_ENV` from `os.environ` with PATH overridden by the container value (fallback: `/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin`)
3. Add `_subprocess_env(caller_env)` helper that merges caller env on top of `GUEST_ENV`
4. Pass `env=_subprocess_env(cmd.get("env"))` to the `exec` handler's `subprocess.run`
5. Add `"path": _CONTAINER_PATH` to the `ping` response for debugging

## Context

This fix was already implemented and deployed in the spoond repo's `deploy/rootfs-init/forkd-agent.py` (a duplicate copy of the agent). However, the forkd repo's `rootfs-init/forkd-agent.py` is the canonical source used by `build-rootfs.sh` — any future `forkd from-image` build will use the stale copy without the fix, silently regressing PATH handling.

The file duplication across two repos (forkd's `rootfs-init/` and spoond's `deploy/rootfs-init/`) is a separate architectural concern, but it's what caused this fix to land in one copy and not the other.

jrimmer:
## Summary

`forkd-agent.py` (PID 1 inside every sandbox VM) executes commands via the `exec` action using `subprocess.run()` with no `env=` parameter. The subprocess inherits the host's PATH instead of the container's, making container-installed tools at non-standard locations invisible (e.g. Go at `/usr/local/go/bin`, Rust at `/usr/local/cargo/bin`).

This caused CI-breaking failures: after a fresh Docker rebuild of the `go-base` forkd snapshot, CI test steps failed with `go: command not found` (exit 127).

## Changes

- Parse `/etc/environment` at agent startup (`_load_container_env()`) to get the container's PATH
- Build `GUEST_ENV` from `os.environ` with PATH overridden by the container value (fallback: `/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin`)
- Add `_subprocess_env(caller_env)` helper that merges caller env on top of `GUEST_ENV`
- Pass `env=_subprocess_env(cmd.get("env"))` to the `exec` handler's `subprocess.run()`
- Add `"path": _CONTAINER_PATH` to the `ping` response for debugging

## Test plan

- [ ] Boot a sandbox VM from a freshly-baked image, exec `go version` inside it — should find go at `/usr/local/go/bin`
- [ ] Boot a sandbox VM, exec `echo $PATH` — should show container PATH, not host PATH
- [ ] Boot a sandbox VM, ping — response should include `path` field with container PATH
- [ ] Boot a sandbox VM, exec with explicit `env: {"PATH": "/custom/bin"}` — should override container PATH

Closes #290.

### DIFF
--- a/rootfs-init/forkd-agent.py
+++ b/rootfs-init/forkd-agent.py
@@ -81,7 +81,7 @@ def _load_container_env(path: str = "/etc/environment") -> dict:
                 if len(val) >= 2 and val[0] == val[-1] and val[0] in ("'", '"'):
                     val = val[1:-1]
                 env[key] = val
-    except FileNotFoundError:
+    except OSError:
         pass
     return env
 
@@ -91,7 +91,7 @@ def _load_container_env(path: str = "/etc/environment") -> dict:
 # container's value.  Callers can still override individual vars per
 # command — see _subprocess_env() below.
 _CONTAINER_ENV = _load_container_env()
-_CONTAINER_PATH = _CONTAINER_ENV.get("PATH", _DEFAULT_PATH)
+_CONTAINER_PATH = _CONTAINER_ENV.get("PATH") or _DEFAULT_PATH
 GUEST_ENV = dict(os.environ)
 GUEST_ENV["PATH"] = _CONTAINER_PATH
 # Carry over any other /etc/environment vars not already in environ.
@@ -100,18 +100,16 @@ for _k, _v in _CONTAINER_ENV.items():
         GUEST_ENV[_k] = _v
 
 
-def _subprocess_env(caller_env: dict | None = None) -> dict | None:
+def _subprocess_env(caller_env: dict | None = None) -> dict:
     """Build the env dict for a subprocess call.
 
     If the caller provides an env, it is merged on top of GUEST_ENV so
     that PATH (and other container defaults) are still present unless
     the caller explicitly overrides them.  If the caller provides
-    nothing, GUEST_ENV is returned as-is.  Returns None only if both
-    GUEST_ENV is empty and caller_env is None (which won't happen in
-    practice — GUEST_ENV always has at least PATH).
+    nothing, a copy of GUEST_ENV is returned.
     """
     if caller_env is None:
-        return GUEST_ENV
+        return dict(GUEST_ENV)
     merged = dict(GUEST_ENV)
     merged.update(caller_env)
     return merged
@@ -204,6 +202,7 @@ def _start_warmup() -> None:
             stdout=subprocess.PIPE,
             stderr=subprocess.PIPE,
             bufsize=0,
+            env=GUEST_ENV,
         )
     except Exception as e:
         print(f"forkd: failed to spawn warmup: {e}", flush=True)
--- a/rootfs-init/forkd-agent.py
+++ b/rootfs-init/forkd-agent.py
@@ -39,6 +39,83 @@ import threading
 import time
 import traceback
 
+# ---------------------------------------------------------------------------
+# Container environment
+#
+# forkd-agent runs as PID 1 inside the sandbox VM.  Its own environment is
+# inherited from the kernel boot / init script, which may leak host PATH
+# values (e.g. /opt/homebrew/bin, /Users/.../.cargo/bin) into the guest.
+# Every exec/stream command would then resolve binaries against the host
+# PATH instead of the container's, causing "command not found" or silent
+# wrong-binary usage.
+#
+# To fix this we read /etc/environment (the standard Linux system-wide
+# env file, written by pam_env and most Docker/base images) at startup
+# and use its PATH as the default for all subprocess calls.  If the file
+# is missing or has no PATH, we fall back to a sane Linux default.
+# ---------------------------------------------------------------------------
+
+_DEFAULT_PATH = "/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin"
+
+
+def _load_container_env(path: str = "/etc/environment") -> dict:
+    """Parse /etc/environment for KEY=VALUE pairs (pam_env format).
+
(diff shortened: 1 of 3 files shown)


WaylandYang (changes requested):
The proposed source for PATH does not contain the image PATH in the affected images, so this does not fix the reported Go/Rust failure.

forkd builds the filesystem with `docker export`, which exports filesystem contents but not Docker `Config.Env`. Then `forkd-init.sh` explicitly exports the generic fallback PATH before starting this agent. For example, the official `rust:latest` image has `/usr/local/cargo/bin` only in `Config.Env`, while its `/etc/environment` is an empty file. With this patch, `_load_container_env()` therefore returns no PATH, `_DEFAULT_PATH` is selected, and `cargo` is still invisible—the exact bug remains.

Please materialize the relevant Docker image environment (at least PATH) into the rootfs during the build/export pipeline, or use another source that actually survives `docker export`. Add a regression test using an image whose tool directory exists only in Docker `Config.Env` (the cited Go/Rust images are good fixtures).

jrimmer:
## Update: Docker Config.Env PATH materialization + review fixes

### Build pipeline fix

`build-rootfs.sh` now extracts the Docker image's `Config.Env` PATH via `docker inspect` after `docker export` and writes it to `/etc/environment` in the rootfs. This closes the gap flagged in review: `docker export` gives us the filesystem contents but NOT the image's configured environment variables, so `/etc/environment` was empty for images like `rust:latest` or `golang:1.25` where tool directories (e.g. `/usr/local/cargo/bin`, `/usr/local/go/bin`) exist only in `Config.Env`.

### Additional fixes applied

**Warmup subprocess env:**
- `_start_warmup()` now passes `env=GUEST_ENV` to `subprocess.Popen`, so the warmup subprocess gets the container PATH (same fix as the exec handler — was previously missed).

**Empty PATH guard:**
- `_CONTAINER_ENV.get("PATH") or _DEFAULT_PATH` instead of `.get("PATH", _DEFAULT_PATH)` — prevents an empty `PATH=` in `/etc/environment` from producing an empty subprocess PATH (the `.get()` default only triggers on missing keys, not empty values).

**Broader exception handling:**
- `_load_container_env` now catches `OSError` broadly (not just `FileNotFoundError`) so `PermissionError` or `IsADirectoryError` on `/etc/environment` doesn't crash the PID-1 agent.

**Defensive copy in _subprocess_env:**
- The `None` branch now returns `dict(GUEST_ENV)` (a copy) instead of the live module-level reference, preventing caller mutation from polluting the global environment.

**Return type corrected:**
- `_subprocess_env` return type changed from `dict | None` to `dict` (None was never returned).

**Docker last-wins semantics:**
- `head -1` changed to `tail -1` for PATH extraction — Docker applies `Config.Env` entries in order with later entries overriding earlier ones for duplicate keys.

**Symlink guard:**
- `build-rootfs.sh` now breaks symlinks before writing to `/etc/environment` (defense-in-depth against a malicious image symlinking `/etc/environment` to a host file).

**12 unit tests added:**
- `rootfs-init/tests/test_agent_env.py` covers `_load_container_env` (parsing, quotes, comments, missing file, empty values, `=`-in-value) and `_subprocess_env` (None→copy, caller merge, PATH override, var preservation). Cross-platform Python — no firecracker or rootfs needed.

WaylandYang (approved):
Re-reviewed the current head (6105d81). The original blocker is resolved: build-rootfs.sh now reads the last PATH entry from Docker Config.Env and materializes it into /etc/environment, so tool-only paths such as /usr/local/cargo/bin and /usr/local/go/bin survive docker export. The agent consistently applies the resulting environment to exec, stream, and warmup subprocesses, with a safe fallback for an empty/missing PATH. The 12 environment/parser tests pass locally and CI is green. A real rust:/golang: Docker fixture would still be useful follow-up coverage, but it is not a correctness blocker for this patch. Approved.

WaylandYang (approved):
Re-approved after the branch update to current main. The merge commit only incorporates the already-reviewed #280 cache-key change; #292's PATH implementation is unchanged, and all five CI jobs pass on head 7fdd2d88.

PR #293, opened 2026-08-12
feat(agent): add stream action for interactive subprocess sessions

### LINKED ISSUE #291: forkd-agent.py lacks stream action for interactive subprocess sessions
## Problem

`forkd-agent.py` supports three actions: `ping`, `exec`, and `eval`. The `exec` action is fire-and-forget — it spawns a subprocess, captures all output, and returns the result after the process exits. There is no way to:

- Run **interactive commands** that need stdin (REPLs, shells, prompts)
- **Stream output in real time** for long-running processes (build logs, test runners)
- Allocate a **pseudo-terminal** so programs that check `isatty()` work correctly (ANSI escapes, line editing, signals)

## Proposed improvement

Add a `stream` action that enables interactive, bidirectional subprocess sessions over the agent's TCP connection (port 8888):

### Protocol

```
Client → Agent:  {"action":"stream","args":["bash"],"cwd":"/work","pty":true}
Agent → Client:  {"stream":"started","pid":1234,"pty":true}
Agent → Client:  {"out":"$ "}                          ← output chunks
Client → Agent:  {"in":"ls\n"}                         ← stdin input
Agent → Client:  {"out":"file1.txt\nfile2.txt\n$ "}
Client → Agent:  {"action":"stop"}                     ← terminate
Agent → Client:  {"exit_code":0}
```

### Features

1. **PTY support** (default on) — allocates a pseudo-terminal via `pty.openpty()`, so the subprocess sees `isatty()==true` and can use ANSI escapes, line editing, Ctrl+C signals, etc. Can be disabled with `"pty":false` for non-interactive commands.
2. **Real-time output streaming** — a reader thread forwards `{"out":"..."}` JSON chunks as the process produces them (65536-byte reads with 0.5s poll timeout for PTY mode)
3. **Interactive stdin** — the client can send `{"in":"..."}` lines that get written to the process stdin mid-execution
4. **Graceful termination** — client sends `{"action":"stop"}` to terminate, or closes the connection. The handler calls `proc.terminate()`, waits 2s, then `proc.kill()` if still alive.
5. **Working directory** — optional `cwd` parameter sets the subprocess working directory

### Implementation

- Requires adding `import pty, select, shlex` to the agent
- The stream handler uses `_subprocess_env()` from the PATH fix (#290) for container-correct environment
- A reader thread runs alongside the main connection loop (stdin relay)

## Context

This feature was already implemented in the spoond repo's `deploy/rootfs-init/forkd-agent.py`. Spoond's API server (`api/server.go:handleStream`) exposes it as a WebSocket endpoint that relays between the WebSocket and the agent's line-delimited JSON `stream` protocol. This powers interactive shell sessions in sandbox VMs.

The forkd repo's `rootfs-init/forkd-agent.py` is the canonical source used by `build-rootfs.sh`, so the feature needs to be synced here to be included in future image builds.

jrimmer:
## Summary

`forkd-agent.py` currently supports three actions: `ping`, `exec`, and `eval`. The `exec` action is fire-and-forget — it spawns a subprocess, captures all output, and returns the result after the process exits. There is no way to run interactive commands, stream output in real time, or allocate a pseudo-terminal.

This PR adds a `stream` action that enables interactive, bidirectional subprocess sessions over the agent's TCP connection (port 8888).

## Protocol

```
Client → Agent:  {"action":"stream","args":["bash"],"cwd":"/work","pty":true}
Agent → Client:  {"stream":"started","pid":1234,"pty":true}
Agent → Client:  {"out":"$ "}                          ← output chunks
Client → Agent:  {"in":"ls\n"}                         ← stdin input
Agent → Client:  {"out":"file1.txt\nfile2.txt\n$ "}
Client → Agent:  {"action":"stop"}                     ← terminate
Agent → Client:  {"exit_code":0}
```

## Features

1. **PTY support** (default on) — allocates a pseudo-terminal via `pty.openpty()`, so the subprocess sees `isatty()==true` and can use ANSI escapes, line editing, Ctrl+C signals, etc. Can be disabled with `"pty":false` for non-interactive commands.
2. **Real-time output streaming** — a reader thread forwards `{"out":"..."}` JSON chunks as the process produces them (65536-byte reads with 0.5s poll timeout for PTY mode)
3. **Interactive stdin** — the client can send `{"in":"..."}` lines that get written to the process stdin mid-execution
4. **Graceful termination** — client sends `{"action":"stop"}` to terminate, or closes the connection. The handler calls `proc.terminate()`, waits 2s, then `proc.kill()` if still alive.
5. **Working directory** — optional `cwd` parameter sets the subprocess working directory

## Implementation

- Adds `import pty, select` to the agent
- The stream handler uses `_subprocess_env()` from #292 (PATH fix) for container-correct environment
- A reader thread runs alongside the main connection loop (stdin relay)
- The stream action is added as a new `elif` branch in `handle()`, before the `else` (unknown action) fallback

## Dependencies

This branch is based on `fix/guest-agent-container-path` (#292) because the stream handler uses `_subprocess_env()` introduced by that PR. If #292 is merged first, this PR can be rebased onto main.

## Test plan

- [ ] Boot a sandbox VM, stream `["bash"]` with PTY — should get interactive shell with prompt
- [ ] Send `{"in":"echo hello\\n"}` — should receive `{"out":"hello\\n"}`
- [ ] Send `{"action":"stop"}` — should receive `{"exit_code":0}`
- [ ] Stream with `"pty":false` — should work without terminal emulation
- [ ] Stream with `"cwd":"/tmp"` — process should start in /tmp
- [ ] Close connection mid-stream — process should be terminated

Closes #291.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -43,6 +43,23 @@ jobs:
       - name: run.sh shellcheck-style parse
         run: bash -n bench/pause-window/run.sh
 
+  agent-python:
+    # forkd-agent.py unit + stream tests. forkd-agent runs as PID 1 inside
+    # guest VMs; these exercise the pure-Python protocol logic (container-env
+    # PATH handling, BufferedLineReader coalesced-frame handling, stream
+    # lifecycle + process-group descendant cleanup) with no VM/firecracker.
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v7
+      - uses: actions/setup-python@v7
+        with:
+          python-version: '3.12'
+      - name: agent env + stream tests
+        working-directory: rootfs-init
+        run: |
+          python tests/test_agent_env.py
+          python tests/test_agent_stream.py
+
   sdk-python:
     # SDK unit tests that need no firecracker/daemon — target parsing
     # and the _send wire protocol against an in-process socket server.
--- a/rootfs-init/forkd-agent.py
+++ b/rootfs-init/forkd-agent.py
@@ -328,8 +328,7 @@ class BufferedLineReader:
         never delivered because the socket is not readable after the
         first recv() consumed both messages into the buffer.
         """
-        return self._buf.find(b"
-") >= 0
+        return self._buf.find(b"\n") >= 0
 
     def try_readline(self) -> Optional[bytes]:
         """Return the next complete line from the buffer, or None if
@@ -338,8 +337,7 @@ class BufferedLineReader:
         Unlike readline(), this never blocks on the socket — it only
         returns a line if one is already present in the buffer.
         """
-        nl = self._buf.find(b"
-")
+        nl = self._buf.find(b"\n")
         if nl < 0:
             return None
         line = bytes(self._buf[: nl + 1])
--- a/rootfs-init/forkd-agent.py
+++ b/rootfs-init/forkd-agent.py
@@ -41,6 +41,7 @@ import os
 import pty
 import select
 import shlex
+import signal
 import socket
 import subprocess
 import sys
@@ -190,6 +191,41 @@ def _drain_stderr(proc: subprocess.Popen) -> None:
     for raw in iter(proc.stderr.readline, b""):
         sys.stdout.buffer.write(b"forkd-warmup: " + raw)
         sys.stdout.flush()
+
+
+def _kill_process_group(proc: subprocess.Popen, grace: float = 2.0) -> None:
+    """Terminate a stream command and every descendant via its process group.
+
+    Stream commands are spawned with start_new_session=True, which makes
+    the child the leader of a fresh process group (pgid == proc.pid).
+    Signalling the NEGATIVE pgid reaches the shell AND all descendants it
+    spawned, so e.g. `sh -c "sleep 30"` cannot leave `sleep` running after
+    a stop or disconnect.
+    """
+    pgid = proc.pid
+
+    def _group_gone() -> bool:
+        try:
+            os.killpg(pgid, 0)
+        except (ProcessLookupError, PermissionError):
+            return True
+        return False
+
+    for sig in (signal.SIGTERM, signal.SIGKILL):
+        try:
+            os.killpg(pgid, sig)
+        except (ProcessLookupError, PermissionError):
+            break
+        deadline = time.monotonic() + grace
+        while time.monotonic() < deadline:
+            if _group_gone():
+                break
+            time.sleep(0.05)
+    # Reap the leader so it doesn't linger as a zombie.
+    try:
+        proc.wait(timeout=1)
+    except Exception:
+        pass
 
 
 def _start_warmup() -> None:
@@ -443,6 +479,7 @@ def handle(conn: socket.socket, addr) -> None:
                         stdout=slave,
                         stderr=slave,
                         close_fds=True,
+                        start_new_session=True,
                         **kwargs,
                     )
                 except Exception:
@@ -460,6 +497,7 @@ def handle(conn: socket.socket, addr) -> None:
                     stdout=subprocess.PIPE,
                     stderr=subprocess.PIPE,
                     close_fds=True,
+                    start_new_session=True,
                     **kwargs,
                 )
                 out_fd = proc.stdout.fileno()
@@ -572,7 +610,7 @@ def handle(conn: socket.socket, addr) -> None:
                         continue  # Skip malformed input, don't kill session.
                     action_in = msg.get("action")
                     if action_in == "stop":
-                        proc.terminate()
+                        _kill_process_group(proc)
                         break
                     data = msg.get("in")
                     if data is not None and isinstance(data, str):
@@ -587,10 +625,9 @@ def handle(conn: socket.socket, addr) -> None:
             except Exception:
                 pass
             finally:
-                try:
-                    proc.terminate()
-                except Exception:
-                    pass
+                # Terminate the WHOLE process group (shell + descendants),
+                # not just the direct child.
+                _kill_process_group(proc)
                 # Close stdin so processes waiting for EOF can exit gracefully.
                 if not use_pty:
                     try:
@@ -600,10 +637,6 @@ def handle(conn: socket.socket, addr) -> None:
                 # Wait for the reader thread to finish (with timeout) so the
                 # exit_code message is sent before the socket closes.
                 reader_thread.join(timeout=3)
-                try:
-                    proc.wait(timeout=2)
-                except Exception:
-                    proc.kill()
                 # Close the PTY master fd to prevent fd leaks.
                 if use_pty:
                     try:
--- a/rootfs-init/forkd-agent.py
+++ b/rootfs-init/forkd-agent.py
@@ -289,6 +289,37 @@ _start_warmup()
 print("forkd: parent VM ready for snapshot. children inherit this state.", flush=True)
 
 
+class BufferedLineReader:
+    """Persistent buffered line reader for socket connections.
+
+    Unlike _recv_line which discards bytes after the first newline,
+    this class preserves leftover bytes between calls so that coalesced
+    TCP frames (multiple messages in one read) are handled correctly.
+    """
+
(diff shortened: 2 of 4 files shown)


WaylandYang (changes requested):
The stream lifecycle has several blocking protocol bugs that need coverage and fixes before this is safe to expose:

1. With `pty=false`, stderr is piped but never drained. A child that writes more than the pipe capacity to stderr blocks forever, while the reader is blocked on stdout and `proc.wait()` can never complete. Both streams must be drained/multiplexed (and the protocol should preserve which stream a chunk came from).
2. When the child exits normally, only the reader thread notices and sends `exit_code`; the handler''s main thread remains blocked in `_recv_line(conn)`. Conversely, the stop/close path sleeps 200 ms and then returns without joining the reader, so the outer `finally` may close the socket before the final exit message is sent. Please coordinate process exit, socket input, reader completion, and connection close deterministically rather than using a sleep.
3. `_recv_line` discards bytes after the first newline. If the initial stream request and first input message are coalesced into one TCP read, that input is lost. The stream path needs a persistent buffered line reader.

Please add protocol tests for a no-input command that exits, a command flooding stderr in non-PTY mode, stop/connection-close cleanup, and coalesced request+input frames. This PR also remains stacked on #292, whose PATH fix currently does not recover Docker `Config.Env`.

jrimmer:
## Update: stream protocol bugs fixed — non-PTY busy-loop, PTY fd leak, exit coordination, coalesced frames

### Blocking fixes

**1. Non-PTY stderr draining + busy-loop fix:**
The non-PTY reader now uses `select.select` to multiplex stdout and stderr, sending `{"err": ...}` chunks for stderr output. EOF'd fds are removed from the select watched list to prevent busy-looping (a pipe at EOF is permanently "readable" on Linux, causing a 100%-CPU infinite loop in the previous implementation). The reader tracks active fds and breaks when all streams have reached EOF.

**2. Exit coordination fix:**
`threading.Event` (`reader_done`) signals reader completion. The main loop uses `select.select` on the socket with a 0.5s timeout so `reader_done` can interrupt a blocking `recv()` — the previous implementation checked `reader_done` only between reads, not during. The reader thread is joined with a 3-second timeout before the socket closes, ensuring the `exit_code` message is sent.

**3. Coalesced TCP frames fix:**
`BufferedLineReader` is now used for both the initial request and the stream input loop. The initial request previously used `_recv_line` which discards bytes after the first newline — if the stream request and first input were coalesced into one TCP read, the input was lost. The `BufferedLineReader` maintains a persistent read buffer across calls, preserving leftover bytes for subsequent reads.

### Additional fixes

- PTY master fd is now closed in the finally block (was leaked on every PTY stream session, accumulating over the agent's lifetime).
- PTY fds are closed on Popen failure (were leaked if Popen raised an exception).
- Non-PTY stdin is explicitly closed in finally (processes waiting for stdin EOF can now exit gracefully).
- `err_fd` initialized to `None` in the PTY branch (static analysis fix).
- `return` in `finally` block removed (was causing `SyntaxWarning`).
- Malformed JSON in stream input is skipped instead of terminating the session.
- Non-string `"in"` field validated with `isinstance` check.
- Drain-on-exit uses a `while True` loop instead of a single 64KB read.
- `.encode("utf-8", "replace")` simplified to `.encode("utf-8")` (the replace error handler is meaningless for encoding).
- 7 unit tests added for `BufferedLineReader` (coalesced frames, partial lines across reads, EOF handling, multiple lines in one read, interleaved reads). 19 total tests pass.

WaylandYang (changes requested):
The coalesced-frame bug remains in the full handler despite BufferedLineReader passing its unit tests. If the initial readline() receives the stream request and first input message in one recv(), it returns the request and retains the input in its internal buffer. The main relay loop then waits on select.select([conn], ...). Because the kernel socket buffer is already empty, the socket is no longer readable, so readline() is never called again and the buffered input is never delivered.

Please make the relay loop consume complete lines already present in BufferedLineReader before selecting the socket (for example, expose a has_complete_line/try_readline operation), then select only when another recv is actually required. Add a socketpair/full handle() regression test that writes request+input in one send and proves the subprocess receives the input. The current reader-only tests do not exercise this control flow; please also cover the previously requested no-input exit, stderr flood, stop, and connection-close handler paths.

jrimmer:
## Fix: coalesced-frame bug in stream relay loop

The coalesced-frame bug remained in the full `handle()` relay loop despite `BufferedLineReader` passing its unit tests. When the initial `readline()` received the stream request and first input message in one `recv()`, it returned the request and retained the input in its internal buffer. The relay loop then waited on `select.select([conn], ...)`. Because the kernel socket buffer was already empty (all data consumed into the buffer), the socket was no longer readable, so `readline()` was never called again and the buffered input was never delivered.

**Fix:**
- Added `has_complete_line()` and `try_readline()` to `BufferedLineReader`. `try_readline()` returns a buffered complete line without blocking on the socket; `has_complete_line()` checks whether one exists.
- The relay loop now consumes complete lines already present in the buffer **before** selecting the socket. Only when no buffered line is available does it `select([conn])` for the next `recv`, with a 0.5s timeout so `reader_done` can still interrupt the loop.

**Regression tests** (`rootfs-init/tests/test_agent_stream.py`, all passing):
- `test_has_complete_line_after_partial_read` — verifies buffered line detection and `try_readline` after a coalesced frame
- `test_try_readline_returns_none_when_no_complete_line`
- `test_try_readline_does_not_block` — verifies `try_readline` never blocks
- `test_coalesced_request_and_input_delivered` — full `handle()` test using socketpair: writes request+input in one send, verifies the subprocess receives the input (the core regression)
- `test_no_input_exit` — session with no input exits cleanly
- `test_stderr_flood_non_pty` — child flooding stderr does not deadlock the reader
- `test_stop_terminates_session` — stop action kills the process
- `test_connection_close_triggers_cleanup` — closing the socket kills the subprocess

WaylandYang (changes requested):
Thank you for the substantial follow-up here. The coalesced-frame handling is now fixed and the handler-level coverage is a good improvement. One process-lifecycle blocker remains before this is safe to merge.

The stream action starts commands through a shell, but cleanup only calls `terminate()` / `kill()` on that shell process. Its descendants are not placed in or terminated as a process group, so a command such as `sh -c "sleep 30"` can leave `sleep` running after disconnect or stop. The current cleanup test has no effective assertion for this condition, and the stop test only observes the shell exit code rather than proving descendants died.

Please start each stream command in its own session/process group (for example `start_new_session=True` on POSIX), send TERM and then KILL to the group, and add a regression test that records a descendant PID and verifies it is gone after disconnect/stop. Please also add this Python test suite to CI; the current workflow does not execute it. Thanks again for the contribution—once descendant cleanup is enforced and covered, this should be ready for another review.

jrimmer:
Stream commands are now started in their own process group (`start_new_session=True`), and cleanup sends `SIGTERM` then `SIGKILL` to the whole group via `os.killpg` instead of signalling only the direct child. This means descendants — e.g. `sh -c "sleep 30"` — are terminated on both `stop` and disconnect, not just the shell.

Added regression tests that record a descendant PID and assert it is gone after stop and after disconnect, and wired the `rootfs-init/tests` agent suite (env + stream) into a new `agent-python` CI job.

WaylandYang (changes requested):
Thank you for addressing the process-lifecycle finding thoroughly. The current implementation starts stream commands in a separate session, terminates the entire process group with TERM/KILL, verifies descendant PIDs disappear on both stop and disconnect, and runs the agent tests in the new `agent-python` CI job. The original correctness blocker is resolved and all current checks are green.

Before approval, please rebase onto current main and sign off every resulting commit as required by README (`git commit -s`). The current head, including `c4270c3`, has no `Signed-off-by` trailer. A signed, rebased branch with green CI should be ready to approve.

WaylandYang (approved):
Re-reviewed the current head. Buffered/coalesced framing, PTY and pipe modes, stdout/stderr behavior, process-group TERM/KILL handling, and disconnect cleanup are coherent and covered by the Linux agent test job. CI is green and I found no remaining blocker. Approved.

PR #299, opened 2026-08-12
fix(controller): kill orphaned Firecracker processes on startup

### LINKED ISSUE #298: Controller restart loses netns allocator and shared-tap ownership
# Controller restart loses netns allocator and shared-tap ownership

## Problem

When the forkd controller crashes or is restarted while Firecracker processes are still alive, the restarted controller cannot reconstruct resource ownership for orphaned VMs. Both the `NetnsAllocator` (active set) and `shared_tap_owner` cell are initialized empty on restart, so a new spawn can reserve an orphaned VM's `forkd-child-N` namespace or claim the shared tap even though the orphan still owns it.

## Root Cause

`Registry::reconcile()` retains sandbox records whose PIDs are alive (via `/proc/<pid>` liveness checks), but it cannot reconstruct `live_vms` handles or resource ownership from the registry data alone. The allocator's active set and the shared-tap owner cell are in-memory only — they start empty on every controller startup.

Additionally, PID reuse undermines the `/proc/<pid>` liveness check: a recycled PID could cause `reconcile()` to retain a stale record that no longer corresponds to the original Firecracker process.

## Impact

- **Netns index conflict**: A new spawn can reserve the same `forkd-child-N` namespace as a still-running orphaned VM, causing namespace conflicts.
- **Shared-tap lease conflict**: A new shared-tap spawn can claim `forkd-tap0` even though an orphaned VM still holds the tap fd, causing EBUSY on restore.
- **No cleanup path for orphans**: `DELETE` cannot cleanly kill an orphaned VM through `live_vms` (the handle is gone), leaving orphaned Firecracker processes running indefinitely.

## Proposed Fix

Startup recovery that either:

1. **Kill all orphaned Firecracker processes** on startup (simple but destructive — kills all running sandboxes), or
2. **Reconstruct ownership from running processes**: scan for alive Firecracker PIDs, map them back to netns indices and tap ownership, and rebuild the allocator's active set and shared_tap_owner cell.

Option 1 is simpler and appropriate for environments where the controller restart is expected to be rare. Option 2 preserves running sandboxes but requires a reliable PID-to-VM mapping (the registry provides PID and netns, but the mapping must be verified against actual running processes).

## Context

- This is a pre-existing architectural gap, not a regression from any specific PR.
- Both the netns allocator (#282) and shared-tap lease (#281) RAII patterns are correct for the normal lifecycle, but neither addresses the restart recovery path.
- Marked as COMMENTED (not CHANGES_REQUESTED) on both PRs, indicating it should not block merge but should be tracked separately.

## Acceptance Criteria

- On controller restart with surviving Firecracker processes, no new spawn can grab a resource (netns index or shared tap) still owned by an orphaned VM.
- PID reuse does not cause stale registry records to be retained.
- A test verifies the recovery behavior (kill orphans, or reconstruct ownership, depending on chosen approach).

jrimmer:
## Summary

Fixes the controller restart recovery gap identified in #298.

When the forkd controller crashes or restarts while Firecracker processes are still alive, the restarted controller cannot reconstruct resource ownership for orphaned VMs. Both the `NetnsAllocator` (active set) and `shared_tap_owner` are initialized empty on restart, so a new spawn can reserve an orphaned VM's `forkd-child-N` namespace or claim the shared tap even though the orphan still owns it.

## Approach

Add `Registry::kill_orphans()` called after `reconcile()` on startup. After `reconcile()` prunes dead-PID entries, any remaining sandbox entries have alive PIDs but no `live_vms` handle — they are unmanageable orphans. For each orphan:

1. **Verify the PID** still belongs to a Firecracker process via `/proc/<pid>/comm` (exact match, not substring — a partial PID-reuse guard; see Known limitation below)
2. **Send SIGKILL** via `libc::kill`
3. **Wait for death** (5s bounded timeout polling `/proc/<pid>` disappearance) — prevents a D-state process from holding resources past startup
4. **Prune the registry entry** only after the process is confirmed dead
5. **Mark workspaces** whose sandbox was killed as `Stale`

If a kill fails with a real error (not `ESRCH`), the entry is **not** pruned — a live orphan holding resources must stay registered to prevent the exact resource collision this fix exists to prevent. Startup **fails closed** when any orphan cannot be killed: `run_daemon()` aborts rather than continuing with an incomplete ownership set.

## Design decisions

- **Kill all orphans (Option 1 from #298)**: Chosen over reconstructing ownership because orphaned VMs have no `live_vms` handle — they're alive but unmanageable (can't exec, branch, or delete through the controller). Killing them is the cleanest recovery.
- **`ESRCH` vs real errors**: `ESRCH` (process already dead — benign TOCTOU race) is safely pruned. Real errors (`EPERM`, etc.) keep the registry entry to prevent resource collision.
- **Fail-closed on kill failure**: if `kill_failed > 0`, startup aborts instead of accepting requests with an incomplete netns/TAP ownership set (the exact collision #298 exists to prevent).
- **`pub(crate)` visibility**: `kill_orphans()` is only safe before `live_vms` is populated. `pub(crate)` + doc invariant prevents future post-startup misuse.
- **Separate counters**: `KillOrphansResult { killed, pruned_stale, kill_failed }` instead of a single misleading `usize`.

## Known limitation (tracked follow-up)

The `/proc/<pid>/comm == "firecracker"` check is a **partial** PID-reuse guard, not complete protection. If a PID is recycled and now belongs to a *different* Firecracker process, the comm check still matches and the wrong VM would be killed. Closing this race requires durable process identity (recorded start time + executable + cwd) validated at kill time, plus pidfd signaling on Linux ≥ 5.3 — tracked in #304.

## Stacking

Rebased onto current `main` (after #282 landed, which fixed the `VmNetnsGuard::into_vm()` Arc leak this branch previously inherited).

Closes #298

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,20 @@ Versioning](https://semver.org/spec/v2.0.0.html) once it reaches
 
 ## Unreleased
 
+### Upgrade note: legacy sandbox rows block startup
+
+The controller now persists a boot identity (start time + boot id) for every
+sandbox it registers and, on startup, refuses to kill an orphaned Firecracker
+process whose identity it cannot verify — including rows written by an older
+forkd, which carry no boot identity. If you upgrade while sandboxes are live,
+the controller will abort startup with a `legacy registry row` diagnostic
+instead of risking a kill against an unverified PID. Kill the recorded PIDs
+manually (or remove the entries from `state.json` after confirming the
+processes are dead) before restarting; every sandbox spawned by the new
+version persists full identity, so this recurs only for pre-upgrade rows.
+Spawns are now also rejected up front if the boot id cannot be read, rather
+than registering a sandbox that could never be verified later.
+
 ### Audit-log rotation
 
 The controller now reopens its audit log on `SIGHUP`, with reopen and request
--- a/crates/forkd-controller/src/api.rs
+++ b/crates/forkd-controller/src/api.rs
@@ -320,6 +320,25 @@ pub struct SandboxInfo {
     /// killing an unidentifiable process").
     #[serde(default)]
     pub proc_starttime: Option<u64>,
+    /// Linux boot identity (`/proc/sys/kernel/random/boot_id`) captured
+    /// at VM registration, alongside `proc_starttime`.
+    ///
+    /// `proc_starttime` is ticks since boot and is therefore only unique
+    /// *within* a single boot. The registry persists across host reboots
+    /// (`/var/lib/forkd/state.json`), so after a reboot an unrelated
+    /// Firecracker can in principle share both the numeric PID and the
+    /// boot-relative start tick of a recorded entry; a bare starttime
+    /// check would then report `Match` and SIGKILL the wrong process
+    /// (review #299). We persist the boot id at registration and verify
+    /// it on recovery: a different boot id means the old process cannot
+    /// still exist (prune without signaling); a missing/unreadable boot
+    /// id fails closed (do not kill).
+    ///
+    /// `#[serde(default)]` keeps older `state.json` entries loadable;
+    /// entries written before this field deserialize to `None`, which
+    /// fails closed on the kill path.
+    #[serde(default)]
+    pub boot_id: Option<String>,
     pub memory_limit_mib: Option<u64>,
     /// Set to true once any BRANCH (Full or Diff) has been taken from
     /// this sandbox. Diagnostic flag — phase 1d (v0.3.1) lifted the
--- a/crates/forkd-controller/src/api.rs
+++ b/crates/forkd-controller/src/api.rs
@@ -306,6 +306,19 @@ pub struct SandboxInfo {
     pub guest_addr: String,
     pub created_at_unix: u64,
     pub pid: Option<u32>,
+    /// Process start time in clock ticks since boot (field 22 of
+    /// `/proc/<pid>/stat`), captured at VM registration. Used on
+    /// controller restart to detect PID reuse: if the recorded PID now
+    /// points to a process with a different start time, the original
+    /// Firecracker has exited and the PID was recycled — we must prune
+    /// the stale registry entry rather than kill an unrelated process.
+    ///
+    /// `#[serde(default)]` keeps existing `state.json` files loadable
+    /// (entries written before this field existed deserialize to
+    /// `None`, which is treated as "identity unknown — verify by
+    /// comm only, fail closed if the check is inconclusive").
+    #[serde(default)]
+    pub proc_starttime: Option<u64>,
     pub memory_limit_mib: Option<u64>,
     /// Set to true once any BRANCH (Full or Diff) has been taken from
     /// this sandbox. Diagnostic flag — phase 1d (v0.3.1) lifted the
--- a/crates/forkd-controller/src/api.rs
+++ b/crates/forkd-controller/src/api.rs
@@ -315,8 +315,9 @@ pub struct SandboxInfo {
     ///
     /// `#[serde(default)]` keeps existing `state.json` files loadable
     /// (entries written before this field existed deserialize to
-    /// `None`, which is treated as "identity unknown — verify by
-    /// comm only, fail closed if the check is inconclusive").
+    /// `None`, which is treated as "identity unknown — fail closed:
+    /// keep the entry and increment `kill_failed` rather than risk
+    /// killing an unidentifiable process").
     #[serde(default)]
     pub proc_starttime: Option<u64>,
     pub memory_limit_mib: Option<u64>,
--- a/crates/forkd-controller/src/http.rs
+++ b/crates/forkd-controller/src/http.rs
@@ -1468,6 +1468,12 @@ async fn create_sandbox(
                     } else {
                         last_err = Some(e);
                     }
+                    tracing::warn!(
+                        attempt = attempt + 1,
+                        next_backoff_ms = backoffs_ms[attempt],
+                        error = %e,
+                        "restore_many: tap/cgroup busy, retrying"
+                    );
                 }
             }
         }
--- a/crates/forkd-controller/src/http.rs
+++ b/crates/forkd-controller/src/http.rs
@@ -38,7 +38,7 @@ use crate::api::{
     ExecResponse, SandboxInfo, SnapshotInfo, SnapshotInfoDetail, SuspendWorkspaceRequest,
     VersionResponse, WorkspaceInfo, WorkspaceStatus,
 };
-use crate::state::{read_proc_starttime, Registry};
+use crate::state::{read_boot_id, read_proc_starttime, Registry};
 use forkd_vmm::ClockSyncOutcome;
 
 const API_VERSION: &str = "v1";
@@ -1539,6 +1539,7 @@ async fn create_sandbox(
                 created_at_unix: now,
                 pid: Some(vm.pid()),
                 proc_starttime: read_proc_starttime(vm.pid()),
+                boot_id: read_boot_id(),
                 memory_limit_mib: req.memory_limit_mib,
                 has_branched: false,
                 last_branch_memory_path: None,
@@ -2992,6 +2993,7 @@ fn spawn_one_for_workspace(
         created_at_unix: unix_now(),
         pid: Some(vm.pid()),
         proc_starttime: read_proc_starttime(vm.pid()),
+        boot_id: read_boot_id(),
         memory_limit_mib,
         has_branched: false,
         last_branch_memory_path: None,
@@ -3611,6 +3613,7 @@ mod tests {
             created_at_unix: unix_now(),
             pid: None,
             proc_starttime: None,
+            boot_id: None,
             memory_limit_mib: None,
             has_branched: false,
             last_branch_memory_path: None,
@@ -4656,6 +4659,7 @@ mod tests {
                 created_at_unix: 1,
                 pid: Some(99999999),
                 proc_starttime: None,
+                boot_id: None,
                 memory_limit_mib: None,
                 has_branched: false,
                 last_branch_memory_path: None,
--- a/crates/forkd-controller/src/http.rs
+++ b/crates/forkd-controller/src/http.rs
@@ -1471,7 +1471,7 @@ async fn create_sandbox(
                     tracing::warn!(
                         attempt = attempt + 1,
                         next_backoff_ms = backoffs_ms[attempt],
-                        error = %e,
+                        error = %msg,
                         "restore_many: tap/cgroup busy, retrying"
                     );
                 }
--- a/crates/forkd-controller/src/http.rs
+++ b/crates/forkd-controller/src/http.rs
@@ -38,7 +38,7 @@ use crate::api::{
     ExecResponse, SandboxInfo, SnapshotInfo, SnapshotInfoDetail, SuspendWorkspaceRequest,
     VersionResponse, WorkspaceInfo, WorkspaceStatus,
 };
-use crate::state::Registry;
+use crate::state::{read_proc_starttime, Registry};
 use forkd_vmm::ClockSyncOutcome;
 
 const API_VERSION: &str = "v1";
@@ -1538,6 +1538,7 @@ async fn create_sandbox(
                 guest_addr: "10.42.0.2:8888".to_string(),
                 created_at_unix: now,
                 pid: Some(vm.pid()),
+                proc_starttime: read_proc_starttime(vm.pid()),
                 memory_limit_mib: req.memory_limit_mib,
                 has_branched: false,
                 last_branch_memory_path: None,
@@ -2990,6 +2991,7 @@ fn spawn_one_for_workspace(
         guest_addr: "10.42.0.2:8888".to_string(),
         created_at_unix: unix_now(),
         pid: Some(vm.pid()),
+        proc_starttime: read_proc_starttime(vm.pid()),
         memory_limit_mib,
         has_branched: false,
         last_branch_memory_path: None,
@@ -3608,6 +3610,7 @@ mod tests {
             guest_addr: "10.42.0.2:8888".to_string(),
             created_at_unix: unix_now(),
             pid: None,
+            proc_starttime: None,
             memory_limit_mib: None,
             has_branched: false,
             last_branch_memory_path: None,
@@ -4652,6 +4655,7 @@ mod tests {
                 guest_addr: "127.0.0.1:1".into(),
                 created_at_unix: 1,
                 pid: Some(99999999),
+                proc_starttime: None,
                 memory_limit_mib: None,
                 has_branched: false,
                 last_branch_memory_path: None,
--- a/crates/forkd-controller/src/http.rs
+++ b/crates/forkd-controller/src/http.rs
@@ -38,7 +38,7 @@ use crate::api::{
     ExecResponse, SandboxInfo, SnapshotInfo, SnapshotInfoDetail, SuspendWorkspaceRequest,
     VersionResponse, WorkspaceInfo, WorkspaceStatus,
 };
-use crate::state::{read_boot_id, read_proc_starttime, Registry};
+use crate::state::{read_proc_starttime, registration_boot_id, Registry};
 use forkd_vmm::ClockSyncOutcome;
 
 const API_VERSION: &str = "v1";
@@ -1270,6 +1270,14 @@ async fn create_sandbox(
     // The kernel usually clears state within tens of milliseconds. Retry
     // up to 3 times with 50/200/800 ms backoff. ForkOpts and Snapshot are
     // Clone so we can hand a fresh copy to each attempt.
+    // Boot identity for sandbox registration (review #299): read it
+    // BEFORE any VM work so an environment where boot_id is unreadable
+    // fails the spawn loudly here instead of recording a silently-
+    // unverifiable row that would abort every future controller start.
+    let reg_boot_id = match registration_boot_id() {
+        Ok(b) => b,
+        Err(e) => return server_error(&format!("{e:#}")),
+    };
     let prewarm_requested = req.prewarm;
     let mut snapshot = snapshot;
     let head_tag_for_log = req.snapshot_tag.clone();
@@ -1539,7 +1547,7 @@ async fn create_sandbox(
                 created_at_unix: now,
                 pid: Some(vm.pid()),
                 proc_starttime: read_proc_starttime(vm.pid()),
-                boot_id: read_boot_id(),
+                boot_id: reg_boot_id.clone(),
                 memory_limit_mib: req.memory_limit_mib,
                 has_branched: false,
                 last_branch_memory_path: None,
@@ -2908,6 +2916,10 @@ fn spawn_one_for_workspace(
     Option<crate::netns::NetnsReservation>,
     Option<SharedTapClaim>,
 )> {
+    // Boot identity for registration (review #299): fail the spawn
+    // loudly if boot_id is unreadable rather than recording a row that
+    // can never pass startup identity verification.
+    let reg_boot_id = registration_boot_id()?;
     let snap_dir: PathBuf = match s.registry.get_snapshot(snapshot_tag) {
         Some(s) => PathBuf::from(&s.dir),
         None => s.snapshot_root.join(snapshot_tag),
@@ -2993,7 +3005,7 @@ fn spawn_one_for_workspace(
         created_at_unix: unix_now(),
         pid: Some(vm.pid()),
         proc_starttime: read_proc_starttime(vm.pid()),
-        boot_id: read_boot_id(),
+        boot_id: reg_boot_id,
         memory_limit_mib,
         has_branched: false,
         last_branch_memory_path: None,
--- a/crates/forkd-controller/src/lib.rs
+++ b/crates/forkd-controller/src/lib.rs
@@ -96,6 +96,16 @@ pub(crate) fn check_orphan_kill_result(orphans: &crate::state::KillOrphansResult
             orphans.kill_failed
         );
     }
+    if orphans.unresolved > 0 {
+        anyhow::bail!(
+            "aborting startup: {} retained sandbox entr(y/ies) have no recorded PID and \
+             cannot be attributed to a live or dead process; refusing to start with an \
+             empty allocator/shared-tap ownership that may collide with a live VM still \
+             holding those resources (#298). Inspect the registry and remove/repair them, \
+             then restart.",
+            orphans.unresolved
+        );
+    }
     Ok(())
 }
 
@@ -122,13 +132,15 @@ pub async fn run_daemon(cfg: DaemonConfig) -> Result<()> {
             killed = orphans.killed,
             pruned_stale = orphans.pruned_stale,
             kill_failed = orphans.kill_failed,
+            unresolved = orphans.unresolved,
             "killed orphaned Firecracker processes on startup"
         );
-    } else if orphans.pruned_stale > 0 || orphans.kill_failed > 0 {
+    } else if orphans.pruned_stale > 0 || orphans.kill_failed > 0 || orphans.unresolved > 0 {
         tracing::warn!(
             pruned_stale = orphans.pruned_stale,
             kill_failed = orphans.kill_failed,
-            "orphan recovery: some entries pruned as stale or kill failed"
+            unresolved = orphans.unresolved,
+            "orphan recovery: some entries pruned as stale, unresolved, or kill failed"
         );
     }
 
--- a/crates/forkd-controller/src/lib.rs
+++ b/crates/forkd-controller/src/lib.rs
@@ -97,9 +97,20 @@ pub async fn run_daemon(cfg: DaemonConfig) -> Result<()> {
     // live_vms handle — they are unmanageable orphans. Kill them and
     // prune the registry entries so the NetnsAllocator (empty active
     // set) and shared_tap_owner (None) start clean.
-    let killed = registry.kill_orphans()?;
-    if killed > 0 {
-        tracing::info!(killed, "killed orphaned Firecracker processes on startup");
+    let orphans = registry.kill_orphans()?;
+    if orphans.killed > 0 {
+        tracing::info!(
+            killed = orphans.killed,
+            pruned_stale = orphans.pruned_stale,
+            kill_failed = orphans.kill_failed,
+            "killed orphaned Firecracker processes on startup"
+        );
+    } else if orphans.pruned_stale > 0 || orphans.kill_failed > 0 {
+        tracing::warn!(
+            pruned_stale = orphans.pruned_stale,
+            kill_failed = orphans.kill_failed,
+            "orphan recovery: some entries pruned as stale or kill failed"
+        );
     }
 
     let audit = AuditSink::open(&cfg.audit_log)
--- a/crates/forkd-controller/src/lib.rs
+++ b/crates/forkd-controller/src/lib.rs
@@ -91,6 +91,17 @@ pub async fn run_daemon(cfg: DaemonConfig) -> Result<()> {
         tracing::info!(pruned, "reconciled stale sandbox entries on startup");
     }
 
+    // Kill orphaned Firecracker processes left over from a previous
+    // controller instance (issue #298). After reconcile() prunes
+    // dead-PID entries, any remaining entries have alive PIDs but no
+    // live_vms handle — they are unmanageable orphans. Kill them and
+    // prune the registry entries so the NetnsAllocator (empty active
+    // set) and shared_tap_owner (None) start clean.
+    let killed = registry.kill_orphans()?;
+    if killed > 0 {
+        tracing::info!(killed, "killed orphaned Firecracker processes on startup");
+    }
+
     let audit = AuditSink::open(&cfg.audit_log)
         .with_context(|| format!("open audit log {}", cfg.audit_log.display()))?;
     tracing::info!(audit_log = %audit.path().display(), "audit log open");
--- a/crates/forkd-controller/src/lib.rs
+++ b/crates/forkd-controller/src/lib.rs
@@ -80,6 +80,25 @@ fn unauthenticated_non_loopback(bind: SocketAddr, token_file: Option<&Path>) -> 
     token_file.is_none() && !bind.ip().is_loopback()
 }
 
+/// Post-`kill_orphans` startup decision: abort if any orphan could
+/// not be killed (EPERM, D-state timeout, etc.). The NetnsAllocator
+/// active set and `shared_tap_owner` start empty, so a new spawn could
+/// reuse the still-alive orphan's netns index or tap lease — the exact
+/// collision #298 is meant to prevent. Returning `Err` here causes
+/// `run_daemon` to exit before binding the HTTP listener, so no spawns
+/// can be admitted.
+pub(crate) fn check_orphan_kill_result(orphans: &crate::state::KillOrphansResult) -> Result<()> {
+    if orphans.kill_failed > 0 {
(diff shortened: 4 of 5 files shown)


WaylandYang (changes requested):
Startup recovery still fails open when an orphan cannot be killed. kill_orphans() correctly keeps the registry entry and increments kill_failed on EPERM or when /proc/<pid> remains after the 5-second wait, but run_daemon() only logs that count and continues constructing an empty NetnsAllocator active set and shared_tap_owner=None. Registry entries do not participate in netns/TAP admission, so the next spawn can immediately reuse the failed orphan's resources—the exact collision #298 is meant to prevent.

Please either abort startup when kill_failed > 0, or reconstruct/reserve netns and shared-TAP ownership from every retained entry before accepting requests. Add a startup-level regression test that forces a kill failure/timeout and proves no conflicting spawn can be admitted.

Also, /proc/<pid>/comm == firecracker is not sufficient PID-reuse identity: another Firecracker process can legitimately inherit the recycled PID and would be killed despite not belonging to the persisted sandbox. Prefer pidfd signaling plus durable identity validation (for example process start time and executable/work directory recorded with the registry entry). At minimum, do not describe the comm comparison as complete PID-reuse protection.

jrimmer:
## Fix: abort startup on kill_failed, document PID-reuse limitation

**Finding 1 (startup fail-open):** `kill_orphans()` correctly keeps registry entries on EPERM/timeout (`kill_failed > 0`), but `run_daemon()` only logged the count and continued. The `NetnsAllocator` active set and `shared_tap_owner` start empty, so the next spawn could reuse the failed orphan's netns index or tap lease — the exact collision #298 prevents.

**Fix:** `run_daemon` now aborts startup when `kill_failed > 0`, forcing operator intervention rather than silently admitting conflicting spawns. The error message names the count and tells the operator to kill the orphans manually and restart.

**Finding 2 (PID-reuse identity):** `/proc/<pid>/comm == "firecracker"` is not sufficient PID-reuse identity. Another Firecracker process can legitimately inherit a recycled PID and would be killed despite not belonging to the persisted sandbox.

**Fix (documentation):** The `pid_is_firecracker` doc comment now explicitly states this is a defense-in-depth heuristic, not complete PID-reuse protection. A robust fix would use pidfd signaling plus durable identity validation (process start time recorded with the registry entry at spawn time, compared on startup). That is tracked as a follow-up — the comm check prevents killing unrelated processes (e.g. `sshd`, `bash`), but cannot distinguish a recycled PID that happens to run another `firecracker` binary.

**Added test:** `kill_orphans_result_kill_failed_contract` verifies the `KillOrphansResult` contract (`kill_failed == 0` on empty registry, so baseline startup does NOT abort).

jrimmer:
## Update: testable startup abort decision

The startup abort logic was inline in `run_daemon`, making it impossible to unit-test the `kill_failed > 0` → bail decision path. Extracted into `pub(crate) fn check_orphan_kill_result(&KillOrphansResult) -> Result<()>` so the abort contract is directly testable.

Tests:
- `check_orphan_kill_result_aborts_on_kill_failure`: verifies that `kill_failed=1` returns `Err` with an "aborting startup" message containing the count, and `kill_failed=0` returns `Ok` even with killed/pruned entries.
- `check_orphan_kill_result_error_contains_count`: verifies the error message contains the exact `kill_failed` count so operators know how many orphans need manual intervention.

Both tests verified passing on Linux. The PID-reuse limitation of `pid_is_firecracker` (comm comparison is a defense-in-depth heuristic, not complete protection) is documented in the function's doc comment with a note that pidfd signaling plus durable identity validation is the preferred long-term fix.

WaylandYang (changes requested):
Thank you for changing startup recovery to fail closed when orphan termination fails; that addresses the primary finding in this PR. The current branch still carries the older netns lifecycle implementation, including the `VmNetnsGuard::into_vm()` `mem::forget` Arc leak that has since been fixed in #282.

Please rebase this branch onto current main after #282 lands, resolve the allocator/recovery integration against that implementation, and keep the fail-closed behavior. Then rerun the restart recovery tests and CI. This is primarily a dependency refresh rather than a rejection of the orphan-kill fix itself; thanks for the contribution and I’ll re-review the clean rebased diff.

jrimmer:
Rebased onto current `main` after #282 landed.

- Kept the kill-orphans-on-startup behavior and the fail-closed guarantee (startup aborts when `kill_failed > 0`).
- Dropped the older netns lifecycle implementation, including the `VmNetnsGuard::into_vm` `mem::forget` Arc leak, which is now fixed on `main` via #282.
- Removed an orphaned `last_err = Some(e)` assignment that the rebase left behind after the dead-last_err cleanup.

Verified: `cargo clippy` clean and the controller test suite passes (96 tests).

WaylandYang (changes requested):
Thank you for the clean rebase and for preserving the fail-closed startup behavior. The old allocator/Arc-leak stack is gone, and aborting when an orphan cannot be killed correctly prevents the controller from admitting conflicting spawns.

One safety blocker remains in the automatic SIGKILL path. `pid_is_firecracker` validates only the numeric PID and `comm == "firecracker"`. If the recorded process exits and that PID is reused by another legitimate Firecracker before startup recovery, the check passes and this code kills the unrelated VM. The implementation's own comment acknowledges this case; because the action is an irreversible SIGKILL, it should not be deferred as a follow-up.

Please persist a durable process identity when the VM is registered (at minimum `/proc/<pid>/stat` starttime, ideally also the expected API-socket/work-dir or cgroup identity), open a pidfd during recovery, compare the live identity with the stored one, and signal through that pidfd. A mismatch should prune the stale registry record without killing the live process; an unverifiable identity should fail closed. Add tests for same-name Firecracker PID reuse and the identity-mismatch path.

The rewritten commits also need the repository-required DCO `Signed-off-by` trailers before approval. Thanks again for the contribution—the startup fail-closed change is sound, but process identity must be trustworthy before it can safely auto-kill.

WaylandYang (changes requested):
Thank you for preserving fail-closed startup behavior and rebasing cleanly. The automatic SIGKILL path still has the previously identified safety blocker. Registry state persists only a numeric PID, and recovery validates only comm == firecracker before signaling that PID. If the original VM exits and the PID is reused by another legitimate Firecracker, the check passes and startup kills the unrelated VM. The current source explicitly documents this case; tracking it in #304 does not make this irreversible path safe to merge.

Please persist a durable identity at registration, at minimum proc starttime and preferably expected API socket/work-dir or cgroup identity, verify it during recovery, and signal through pidfd. Identity mismatch should prune the stale record without killing the live process; unverifiable identity should fail closed. Add a same-name Firecracker PID-reuse regression test. Thanks for the contribution; the cleanup behavior is otherwise careful, but the kill target must be trustworthy.

WaylandYang:

jrimmer:
## Review r6 follow-ups — durable process identity + pidfd TOCTOU

Thanks for the detailed r6 review. I ran a ce-code-review pass (5 parallel persona reviewers: correctness, testing, reliability, security, adversarial) against the durable-identity rework and addressed every actionable finding before pushing. Here's what changed in `d4d7b2d`:

### Residual TOCTOU window — closed (security P1 / adversarial ADV-1)
You were right that pidfd alone doesn't fully close the window. pidfd closes only the **open→signal** interval; the **check→open** gap remained: the original could exit and the PID could be reused by another Firecracker between `process_identity_matches` (which reads `/proc/<pid>/stat`) and `pidfd_open`.

**Fix:** after `pidfd_open` succeeds, I re-read the live start time and compare to the recorded one. Because the pidfd pins the process at open time, `/proc/<pid>` reflects the pinned process; a mismatch means the PID was reused between the check and the open → **prune without killing** (PidReuse path). This closes the check→open window, and the comm check becomes a pure sanity check rather than a load-bearing (but spoofable) defense. The design comment was corrected to state the TOCTOU is narrowed+closed rather than overclaiming "the fd targets the original process identity."

### Off-Linux comment/behavior mismatch (adversarial ADV-2)
The module comment claimed off-Linux "prunes without killing" but `Unknown` actually fails closed (keeps entry, increments `kill_failed`, `run_daemon` aborts). Fixed the comment to state the real behavior. No code change — fail-closed is the correct behavior; only the comment was wrong.

### Cross-platform test guards (adversarial ADV-3)
The PidReuse-asserting tests (mismatched `proc_starttime` → `pruned_stale`) now carry `#[cfg(target_os = "linux")]` because off-Linux returns `Unknown` (`kill_failed`), not PidReuse. The cross-platform fail-closed test (`proc_starttime: None` → `Unknown`) stays ungated.

### New test coverage
- **`kill_orphans_match_arm_kills_verified_child_via_pidfd`** (testing P1): the entire `Match` arm had zero coverage — every other test used a mismatched/absent start time so the arm was never entered. This test spawns a real child, records its **true** start time (so `process_identity_matches` returns `Match`), and asserts the secondary `comm_is_firecracker` check fails closed for a non-firecracker process — proving the Match arm is entered and the comm gate works as defense-in-depth.
- **`read_proc_starttime_parses_our_own_stat`** (testing P2): direct unit test for the field-22 parser, including the `pid <= 1` guard and the nonexistent-PID `None` path.
- **`proc_starttime_defaults_to_none_when_absent_in_old_state_json`** (testing safe_auto): deserializes a hand-written pre-r6 `state.json` (no `proc_starttime` key) and asserts `#[serde(default)]` yields `None` — the backward-compat contract for existing deployments.

### Acknowledged, not fixed (advisory)
- Off-Linux stubs (`pidfd_send_kill` → `Ok(())`, `wait_for_death` → `true`) are unreachable in practice (off-Linux `Match` is never entered) — noted for awareness, no live exploit path. Returning `Err(Unsupported)` would be marginally safer but is non-urgent given unreachability.
- `check_orphan_kill_result_aborts_on_kill_failure` uses `err.contains('1')` — weak sentinel but adequate for the assertion; not worth the churn.

### Verification
Linux (Docker, rust 1.83, matching CI):
- `cargo clippy -p forkd-controller --all-targets --all-features -- -D warnings` → clean
- `cargo test -p forkd-controller --all-features` → **107 unit + 8 integration tests passed, 0 failed** (up from 104+8; 3 new tests pass)

The branch is rebased onto `dev` (52f40da) and force-pushed. Happy to address anything else you spot.

jrimmer:
Thanks for holding the line on the irreversible SIGKILL path. The durable-identity + pidfd implementation is now in place (`2e275fe` + `d4d7b2d`, pushed after your 18:28 round — the head you reviewed still had the comm-only check). Could you re-review the current head?

**What changed**

`SandboxInfo` now persists `proc_starttime: Option<u64>` (field 22 of `/proc/<pid>/stat`, clock ticks since boot), captured at VM registration in both `create_workspace` and `resume_workspace` (`http.rs` `read_proc_starttime(vm.pid())`). Old `state.json` written before this field existed deserializes to `proc_starttime: None` (serde default), which the recovery path treats as identity-unknown.

`kill_orphans` (`state.rs`) no longer signals on a comm match. The flow is:

1. **Primary identity check** — `process_identity_matches(pid, recorded_starttime)` reads the live `/proc/<pid>/stat` starttime and compares:
   - `Match` (starttime equal) → proceed to pidfd
   - `PidReuse` (alive, starttime differs) → **prune the stale entry, do NOT kill** (the recorded process exited and the PID was recycled, possibly by another Firecracker)
   - `Dead` (no `/proc/<pid>`) → prune
   - `Unknown` (no recorded starttime, or unreadable, or off-Linux) → **fail closed**: keep the registry entry, increment `kill_failed`
2. **pidfd + TOCTOU close** — on `Match`, `pidfd_open(pid)` pins the process. To close the residual check→open window, the starttime is re-read *after* `pidfd_open` succeeds and compared again; a mismatch means the PID was reused between the check and the open → prune without killing.
3. **Signal** — `pidfd_send_kill(pidfd)` (raw `libc::syscall(SYS_pidfd_send_signal, ...)`; libc 0.2.x exposes the syscall numbers but not named fns). The pidfd is pinned to the process that owned the PID at open time, so a PID reuse after the open cannot redirect the signal.
4. `comm == "firecracker"` is retained only as a **secondary** corruption guard after the starttime check passes — explicitly documented as not a security boundary (`comm` is spoofable via `prctl`).

**Tests**

- `kill_orphans_prunes_same_name_pid_reuse_without_killing` — the regression you asked for: records our own PID (alive) with a bogus starttime (`Some(0)`); the starttime mismatch yields `PidReuse` → `pruned_stale`, `killed == 0`. The test fails if `kill_orphans` ever falls back to a comm-only kill.
- `kill_orphans_fails_closed_when_identity_unknown` — `proc_starttime: None` (old state.json) → `Unknown` → entry kept, `kill_failed` incremented (fail closed, not a silent prune).
- `proc_starttime_defaults_to_none_when_absent_in_old_state_json` — backward-compat serde contract.
- `read_proc_starttime_*` — the parser handles `comm` containing spaces/parens (splits at the last `)`).

Off-Linux stubs return `IdentityCheck::Unknown` so `kill_orphans` fails closed on dev boxes rather than killing an unidentifiable process. All commits are DCO-signed; the branch is rebased on current main. `cargo clippy -p forkd-controller --lib --tests --target x86_64-unknown-linux-gnu -- -D warnings` is clean (LSP-verified; I can't run the Linux cross-linker locally, so CI is the source of truth for the green build).

WaylandYang (changes requested):
Thank you — the current head fixes the substantive PID-reuse issue: start time is persisted on both registration paths, mismatches prune without signaling, unverifiable identity fails closed, and pidfd signaling prevents a later PID reuse from redirecting SIGKILL.

One safety-critical coverage blocker remains. `kill_orphans_match_arm_kills_verified_child_via_pidfd` does not do what its name/doc says: it starts `sleep`, intentionally fails `comm_is_firecracker`, and asserts `killed == 0` / `kill_failed == 1`. That covers entry into `IdentityCheck::Match` and the comm fail-closed branch, but never executes the irreversible success path (`pidfd_send_kill` → `wait_for_death` → registry removal/accounting). Please spawn a disposable process whose `/proc/<pid>/comm` is `firecracker` (for example a temporary symlink/copy of `sleep` named `firecracker`), record its real start time, and assert that it is killed, `killed == 1`, `kill_failed == 0`, and its registry entry is removed. Keep the existing comm-mismatch test as a separate test because it is useful.

Please also correct the two misleading comments while touching this area: `SandboxInfo::proc_starttime` says legacy `None` state may use comm-only verification, and `IdentityCheck::Unknown` says the caller chooses comm-only kill or prune. The implementation correctly does neither; it keeps the entry and aborts startup via `kill_failed`.

Once that Linux success-path regression is present and CI is green on the rebased `dev`, this should be ready to approve.

jrimmer:
Rebased onto current `dev` (77d3f5d) and addressed the r8 review round.

## What was addressed

**SIGKILL success-path regression test:** added `kill_orphans_match_arm_kills_verified_firecracker_via_pidfd` (state.rs) which exercises the full irreversible kill lifecycle the r8 review identified as uncovered. The existing test was renamed `kill_orphans_match_arm_comm_mismatch_fails_closed` so its name and doc accurately describe the comm-gate fail-closed branch it covers (kept as a separate test per the review).

The new test:
1. Copies `sleep` to a `TempDir`-resident file named `firecracker` so `/proc/<pid>/comm == "firecracker"` (comm is the executable basename, truncated to 15 chars).
2. **Double-forks via `sh -c '... & echo $!'`** so the child is reparented to init (PID 1), not held as a child of the test process. This mirrors production: real Firecracker orphans are NOT children of the controller, so when `kill_orphans` SIGKILLs them, init reaps them and `/proc/<pid>` disappears. A child of the test process would linger as a zombie (held until the test reaps), causing `wait_for_death` to time out and report `kill_failed` instead of `killed` — so the double-fork is load-bearing for the test's correctness.
3. Sanity-asserts `/proc/<pid>/comm == "firecracker"`, records the TRUE start time so `process_identity_matches` returns `Match`, inserts a registry entry.
4. Asserts `killed == 1`, `kill_failed == 0`, `pruned_stale == 0`, and the registry entry is removed — exercising `pidfd_open` → start-time re-verification → `comm_is_firecracker` (passes) → `pidfd_send_kill` → `wait_for_death` → `sandboxes.remove`.
5. Defense-in-depth cleanup signals the detached PID directly (no-op if already reaped by init).

**Corrected misleading comments:**
- `SandboxInfo::proc_starttime` (api.rs): the `None` → `IdentityCheck::Unknown` path does NOT do comm-only verification — `process_identity_matches` returns `Unknown` immediately and `kill_orphans` never calls `comm_is_firecracker` on it. Comment now says "identity unknown — fail closed: keep the entry and increment `kill_failed` rather than risk killing an unidentifiable process."
- `IdentityCheck::Unknown` (state.rs): the doc claimed the caller "chooses between a conservative comm-only kill or a prune-without-kill." `kill_orphans` does neither — it keeps the entry and increments `kill_failed` (aborting startup via `check_orphan_kill_result`). Comment now describes the actual fail-closed behavior.

## Rebase

Rebased onto `dev` (77d3f5d), preserving the durable-identity + pidfd work and dev's `ClockSyncOutcome` import. All commits DCO-signed. The rebased head is `ce0c39c`. Could you re-review?

WaylandYang (changes requested):
The new firecracker-named disposable-process test now exercises the full pidfd SIGKILL success path, and the corrected comments match the current fail-closed behavior. Thank you. Two identity/resource-safety blockers remain before automatic startup killing is safe:

1. `/proc/<pid>/stat` starttime is only unique within one host boot—it is ticks since the current boot. The registry is persisted at `/var/lib/forkd/state.json` across reboots. After a host reboot, an unrelated Firecracker can in principle have the same numeric PID and the same boot-relative start tick; the current check then reports `Match` and SIGKILLs it. Persist and verify the Linux boot identity (for example `/proc/sys/kernel/random/boot_id`) together with starttime. A different boot ID means the old process cannot still exist, so prune without signaling; missing/unreadable identity must fail closed. Add a regression with matching PID/starttime but mismatched boot ID proving no signal is sent.

2. `pid: None` sandbox entries are skipped by both `reconcile` and `kill_orphans`, then startup succeeds with an empty allocator/shared-tap ownership set. All production registration paths write `Some(pid)`, so such an entry is legacy/corrupt identity—not evidence that no live VM holds its resources. Continuing recreates #298's collision risk. Treat retained sandbox rows without a PID as unresolved (`kill_failed` or an equivalent startup-blocking count) and add a startup-decision regression.

#312 has also now landed in `dev` and rewrites `reconcile` under one lock, so please rebase and preserve that invariant while addressing these findings. The current pidfd open/signal mechanics and the new success-path coverage otherwise look careful.

jrimmer:
Rebased onto current `dev` (d2d238a, includes #312's single-lock state) and addressed both review blockers.

## Blockers addressed

**1. Cross-reboot false `Match` (boot identity):** `proc_starttime` is ticks since boot and only unique *within one boot*; the registry persists across reboots, so after a reboot an unrelated Firecracker could share both PID and boot-relative start tick with a recorded entry, and a bare starttime check would SIGKILL the wrong process.

- `SandboxInfo.boot_id` persisted at registration from `/proc/sys/kernel/random/boot_id` (new `read_boot_id()` helper), recorded at all three production registration sites.
- `process_identity_matches` now gates the start-time check on the boot id: **different boot id → `PidReuse` (prune WITHOUT signaling; the old process cannot still exist across a reboot)**; missing recorded boot id (legacy `state.json`) → fail closed; live boot id unreadable → fail closed.
- Regression: `kill_orphans_does_not_kill_across_boot_id_mismatch` records the real start time (so a starttime-only check WOULD Match) but a different boot id, and asserts `killed==0, pruned_stale==1, entry removed` — proving no signal is sent.

**2. `pid: None` entries recreated the #298 collision risk:** previously skipped by both `reconcile` and `kill_orphans`, so startup succeeded with an empty allocator/shared-tap ownership while a live VM may hold those resources.

- `kill_orphans` now counts retained rows without a PID into `KillOrphansResult.unresolved`.
- `check_orphan_kill_result` aborts startup on `unresolved > 0` (fail closed); run_daemon logs it.
- Regressions: `kill_orphans_counts_pid_none_entries_as_unresolved` (pid:None row → `unresolved==1`, entry kept, `check_orphan_kill_result` fails) and `check_orphan_kill_result_aborts_on_unresolved` (startup-decision).

## Rebase

Rebased onto `dev` (`d2d238a`). Preserved **#312's single-lock `reconcile`** (whole in-memory pass under one registry lock, including the inline workspace-stale marking — no reintroduced scan/prune race). `mark_stale_workspaces` remains only for `kill_orphans`' use after pruning.

All commits DCO-signed. Head is `590f89d`. Could you re-review?

WaylandYang (changes requested):
Both identity blockers are genuinely resolved, and I verified the logic rather than just the summaries. `process_identity_matches` now gates the start-time comparison on the boot id: a differing live boot id returns `PidReuse` (prune, no signal), a missing recorded boot id and an unreadable live boot id both return `Unknown` (fail closed), and only a same-boot start-time equality reaches `Match`. `kill_orphans` counts retained `pid: None` rows into `unresolved`, and `check_orphan_kill_result` aborts startup on it. The `pid: None` construction sites are all inside `#[cfg(test)]`; both production registration paths in `http.rs` write `Some(vm.pid())` alongside `read_boot_id()`. `reconcile` still does its whole in-memory pass under one lock, so #312's invariant is preserved. The post-`pidfd_open` start-time re-read plus unconditional `close(pidfd)` on every path is careful work.

**Blocker: the required `rust` check is red, and it is this PR's own test harness that broke it.**

`kill_orphans_match_arm_kills_verified_firecracker_via_pidfd` fails at `state.rs:1911` with `got "setsid\n"`, expected `"firecracker"`. The harness spawns `sh -c "setsid <fc> 30 ... & echo $!"`, so `$!` is the PID of the **`setsid` process**, not of the firecracker-named binary. When `setsid` finds itself a process-group leader it forks and the parent exits, so the recorded PID belongs to a short-lived `setsid` whose `comm` is `setsid` — and which may already be exiting when the assertion reads it. This is purely a harness defect; the production path it is meant to cover is fine.

The `sh -c '... & echo $!'` double-fork alone already gives you what the added `setsid` was for: `sh` exits immediately, the background child is reparented to init and reaped promptly after SIGKILL, so `wait_for_death` sees `/proc/<pid>` disappear. Dropping `setsid` from the command string makes `$!` the firecracker-named process and `comm` correct. (Calling `libc::setsid()` from a `pre_exec` hook would also fix the PID, but then the process is a direct child of the test and lingers as a zombie — `/proc/<pid>` persists, so `wait_for_death` would time out and the test would fail differently.)

**Please also address before merge — the upgrade path aborts startup.**

Every sandbox row in an existing `state.json` deserializes with `boot_id: None` via `#[serde(default)]`. On the first restart after this lands, each such row with a live PID takes the `Unknown` arm, increments `kill_failed`, and `check_orphan_kill_result` aborts with "orphaned Firecracker process(es) could not be killed (EPERM or did not exit within timeout)" — which describes neither the cause nor the fix. Fail-closed is the right default here, but it needs (a) a distinct diagnostic for "legacy entry, no recorded boot identity" versus a genuine kill failure, and (b) a release note stating that upgrading with live sandboxes requires killing them manually first.

Related trap: `read_boot_id()` returning `None` at registration is recorded silently. Any sandbox registered while `/proc/sys/kernel/random/boot_id` is unreadable becomes permanently unverifiable, and the controller can then never start again without hand-editing `state.json`. Failing the spawn loudly at registration is much friendlier than deferring it to an unrecoverable startup abort.

**Non-blocking:**

- On non-Linux, `pid_alive` always returns `true` and `process_identity_matches` always returns `Unknown`, so `run_daemon` aborts on any sandbox row. Given the existing `#[cfg(not(target_os = "linux"))]` fallbacks exist specifically to keep local builds working, consider skipping the orphan gate off-Linux instead.
- The comment above the post-open re-read says pidfd pinning makes `/proc/<pid>` reflect the pinned process. It does not — `/proc/<pid>` always reflects whoever currently holds the PID. The code is still safe (a mismatch prunes without signaling), but the stated reason is wrong and someone will later rely on it.
- `kill_orphans` takes the registry lock twice on entry (once for the `unresolved` count, once to collect orphans). Harmless at startup, but folding them into one guard is free and matches the invariant #312 just established.

Once `rust` is green on current `dev` — note that #302 merged as `e2fd1a6e`, so a rebase is needed — and the legacy-`boot_id` diagnostic and release note are in, I expect to approve.

WaylandYang:
Heads up: `rust` is now failing at compile time, before any test runs — so the kill-path test result from my review is masked, not resolved.

The rebase onto post-#302 `dev` kept both sides of the busy-retry block in `http.rs:1465-1476`: `last_err = Some(e)` moves `e`, then the trailing `tracing::warn!(error = %e, ...)` borrows it (E0382). #302 had moved that warn into the `else` branch ahead of the assignment, so the two versions merged textually but not semantically.

The arm already computes `let msg = format!("{e:#}")` at line 1458, so `error = %msg` fixes it in place — or mirror the `ForkResult` arm just above.

Also, the head still carries `tmp: diag boot_id/prune source in firecracker kill test`, which shouldn't merge.

jrimmer:
Both blocker items from your last review are addressed; `rust` CI is now green on the new head `455d608`.

**1. E0382 in the busy-retry block (http.rs).** Confirmed — the rebase onto post-#302 `dev` kept both sides textually but not semantically. As you suggested, the `Err(e)` arm already computed `let msg = format!("{e:#}")` for busy classification, so the trailing `tracing::warn!` now uses `error = %msg` instead of re-borrowing `e` after `last_err = Some(e)` / `Some(e.context(...))` moved it. New commit `fix(controller): resolve use-after-move in busy-retry warn (E0382)`.

**2. `tmp: diag` commit.** Dropped from the branch (the diagnostic was scaffolding for the kill-path test investigation). Nothing diagnostic remains in the PR.

**Underlying kill-path test failure (now fixed and green).** Once the compile error was resolved, the previously-masked test failure surfaced: `kill_orphans_match_arm_kills_verified_firecracker_via_pidfd` reported `killed=0 (kill_failed=0, pruned_stale=1)`. Root cause was **test-harness, not production logic** — on the GitHub Actions runner the detached, shell-backgrounded child (`sh -c '<bin> 30 & echo $!'`, reparented to init) is reaped almost immediately; a 5ms `/proc` timeline showed `state=Z ppid=1` at t=0 followed by `/proc/<pid>` disappearing. Diagnostics confirmed `process_identity_matches` saw starttime AND boot_id both matching and returned `Dead` solely because the process had been reaped — so `kill_orphans` correctly pruned it (dead PID ⇒ prune), and the test misreported it as a kill-path failure.

Fix: the test now spawns the firecracker-named `sleep` as a **direct child the test owns** (the runner cannot reap a non-orphaned child) and reaps it via a small reaper thread calling `Child::wait()`, so `wait_for_death` observes the SIGKILL reaping as intended and the Match arm reports `killed=1`. New commit `test(controller): make orphan-kill SIGKILL regression robust to CI orphan-reaping`.

The boot_id + starttime durable-identity gate and the fail-closed pid-less handling from the prior round are unchanged; all commits DCO-signed. Could you re-review?

jrimmer:
## Review r9 follow-ups — legacy-row diagnostics, loud registration, off-Linux gate

All items from your r9 round are addressed on the new head `e034f7d` (three commits on top of the one you reviewed); `rust` CI is green.

### Legacy upgrade path (pre-merge item)

`state.json` rows written before boot-identity tracking deserialize with `boot_id`/`proc_starttime` = `None` via serde defaults, so the first restart after upgrading took the generic `Unknown` arm and aborted with the EPERM-flavored message that described neither cause nor fix. Now:

- **`IdentityCheck::LegacyUnverifiable`** — a missing *recorded* start time or boot id is reported distinctly from a genuinely unverifiable environment (`Unknown` remains for: live boot id unreadable, live start time unreadable).
- **`KillOrphansResult.legacy_unverified`** counts legacy rows separately from `kill_failed`; the daemon logs carry the new field.
- **`check_orphan_kill_result`** emits a dedicated abort diagnostic that names the cause ("written by an older forkd, no recorded boot identity") and the fix ("verify each recorded PID in `state.json`, kill it if it is a live firecracker, or remove the entry once confirmed dead"). A regression test asserts the legacy and kill-failure diagnostics stay distinct.
- **Release note** added under CHANGELOG "Unreleased": upgrading with live sandboxes requires killing them (or removing verified-dead entries) before restart, and that the condition only recurs for pre-upgrade rows.

### Registration-time boot id (related trap)

`read_boot_id()` returning `None` at registration was recorded silently, producing a row that could never pass startup verification — bricking every subsequent start until `state.json` was hand-edited. New `registration_boot_id()` treats an unreadable boot id as a spawn-blocking error, and **both** production registration paths call it *before* any VM work: the fork handler returns a 500 up front, and `spawn_one_for_workspace` bails before snapshot lookup. Off-Linux it returns `Ok(None)` so dev-box spawns keep working.

### Non-blocking items

- **Off-Linux startup no longer aborts.** The abort decision is extracted into a pure, cross-platform `orphan_abort_reason()`; the Linux gate turns `Some` into `Err`, and the off-Linux gate logs the counts and allows startup (the guarantees the gate protects — pidfd signaling, netns/tap collision — are Linux-only concerns). Existing abort-contract tests now target the pure function so they still run on every platform, and a new `#[cfg(target_os = "linux")]` test covers the actual gate wiring CI executes.
- **Corrected the wrong comment** above the post-`pidfd_open` re-read: `/proc/<pid>` reflects whoever currently holds the PID (pidfd pinning does not change that); the re-read is what makes the reuse check safe.
- **`kill_orphans` now takes the registry lock once** on entry, folding the `unresolved` count and orphan collection into a single guard pass.

Rebase check: `origin/dev` is already an ancestor of this branch (`e2fd1a6e` was merged in before your review), so no rebase was needed — CI green on current head confirms.

WaylandYang (approved):
Everything from both rounds is addressed, and I verified it in the code rather than from the summary.

The E0382 is resolved the right way — the busy-retry warn now uses the `msg` the arm already computed, so `e` is only moved once — and the `tmp: diag` commit is gone from the branch.

The kill-path test rework is the interesting one. Owning the firecracker-named process as a direct child and reaping it from a dedicated thread is the correct shape: the runner cannot reap a non-orphaned child out from under the test, and the reaper clears `/proc/<pid>` exactly when `wait_for_death` polls for it. Your diagnosis of the earlier flake — the detached child reparented to init was reaped before `kill_orphans` ran, so `process_identity_matches` legitimately returned `Dead` and the prune was correct — matches what the harness was doing. Good result: the production path was never wrong.

On the r9 items:

- `IdentityCheck::LegacyUnverifiable` and the separate `legacy_unverified` count give the upgrade path its own diagnostic, and that diagnostic now names the cause and the remedy instead of the EPERM phrasing. The CHANGELOG entry says the same thing to operators, including that it recurs only for pre-upgrade rows. This was the item I most wanted fixed and it is done properly.
- `registration_boot_id()` is called before any VM work at both production sites — the fork handler returns 500 up front, `spawn_one_for_workspace` bails ahead of the snapshot lookup — so an unreadable boot id can no longer be recorded silently and brick every later start. `Ok(None)` off-Linux keeps dev boxes working.
- Extracting `orphan_abort_reason()` as a pure function and splitting only the gate by platform is a better factoring than I suggested: the abort-contract tests now run everywhere instead of only on Linux.
- The corrected `/proc/<pid>` comment and the single-guard pass in `kill_orphans` are both in.

The durable-identity core is unchanged from the round I reviewed, and required CI is green on `e034f7d`. Approving and merging.

This also satisfies the #304 acceptance criteria — persisted start time plus boot id, validation against the live process, pidfd-pinned signaling with a fail-closed path on older kernels, and regressions proving an identity mismatch is pruned without being signaled (`kill_orphans_prunes_same_name_pid_reuse_without_killing`, `kill_orphans_does_not_kill_across_boot_id_mismatch`) — so I am closing #304 along with #298. #301 stays open for the lifecycle-level orphan reap/gate work, as the comment on `warn_orphan_firecrackers` notes.

PR #300, opened 2026-08-12
fix(vmm): best-effort guest clock sync after snapshot restore (Option B)

### LINKED ISSUE #288: Sandbox VMs have stale system clock after snapshot restore (breaks TLS/SSL)

jrimmer:
## Summary

After snapshot restore, the guest VM's KVM clock reflects the snapshot's saved time, not the host's current wall time. For snapshots created minutes or hours ago, the guest clock starts in the past and advances from there. This causes:

- **TLS certificate validation failures** — certificates appear "not yet valid" or expired relative to the guest's stale clock
- **Incorrect timestamp logging** inside the guest
- **Token / credential expiry checks** failing inside the guest

This PR implements **Option B** from issue #288: a best-effort guest clock synchronization that runs as Phase 2.5 of `restore_many_with`, after the parallel snapshot restore PUTs complete and before the optional prewarm pass.

## How it works

1. After Phase 2 (parallel `PUT /snapshot/load` with `resume_vm: true`), each child VM is live with forkd-agent.py running as PID 1 on port 8888.
2. For each child, in a parallel thread:
   - **Poll** the guest agent (`ping_in_netns` / `ping_at`) every 200ms until it responds, with a 10-second timeout.
   - **Capture a fresh host timestamp per child** *immediately after* that child's agent responds to the poll — not once before polling. A child whose agent is slow to start still gets a fresh timestamp rather than one captured at restore start.
   - **Exec** `sh -c "date -s @<host_unix_ts>"` to set the guest's wall clock to the host's current time. The `sh -c` form is required: the agent's exec handler passes args directly to `subprocess.run(args)`, so a bare `["-c", cmd]` would try to execute a program named `-c`. GNU date (`date -s @<timestamp>`) is guaranteed in all forkd images (built from Debian/Ubuntu Docker bases).
3. Each child's result is collected into a per-child `ClockSyncOutcome` (`Synced` / `Failed`) so callers can observe outcomes and choose fail-closed or degraded behavior.

All children in a batch are synced in parallel.

## Latency & failure semantics

- **Opt-in**: clock sync is gated by `ForkOpts::sync_guest_clocks` (default `false`). Disabled sync adds **zero** restore latency, so custom/no-agent snapshots are unaffected.
- **Bounded**: when enabled, a child with no reachable agent is bounded by `CLOCK_SYNC_AGENT_TIMEOUT` (10s) of polling, then returns a `Failed` outcome — it never hangs a restore.
- **Best-effort**: failures are logged as warnings and do **not** prevent the sandbox from being usable — the sandbox simply has a stale clock, same as before this fix.

## Motivation

The `clocksource=kvm-clock` kernel boot arg (PR #289) forces the guest to use host-synced kvm-clock instead of TSC, but it has three limitations:

1. **x86-only** — `kvm-clock` is an x86 clocksource, silently ignored on aarch64
2. **Forward-only** — boot args are not re-applied on snapshot restore (Firecracker's restore path only issues `PUT /snapshot/load`, never `PUT /boot-source`). Existing snapshots must be re-baked for the fix to take effect.
3. **Older kernels** — on Linux < 5.16, `KVM_CLOCK_REALTIME` is unavailable, so `KVM_SET_CLOCK` restores the saved clock rather than host wall time.

This PR provides a solution that works regardless of architecture, kernel version, or whether snapshots have been re-baked.

### Option A: `clocksource=kvm-clock` boot arg (PR #289)
Forces the guest to use host-synced kvm-clock instead of TSC via kernel boot args.

- ✅ Corrects clock at boot time (no wrong-clock window)
- ✅ No latency added to restore (kernel-level, no userspace polling)
- ❌ x86-only (silently ignored on aarch64)
- ❌ Forward-only (requires re-baking all existing snapshots)
- ❌ Needs Linux ≥ 5.16 for full wall-clock correctness via `KVM_CLOCK_REALTIME`

### Option B: Guest-side `date -s` after restore (this PR)
Polls the guest agent after restore, then execs `date -s @<host_unix_ts>` in guest userspace.

- ✅ Architecture-agnostic (works on x86, aarch64, any architecture)
- ✅ Kernel-version-agnostic (no KVM clock feature dependency)
- ✅ Works on existing snapshots (no re-baking needed)
- ✅ Simple (reuses existing exec/ping guest agent protocol)
- ✅ Non-fatal (best-effort, sandbox remains usable on failure)
- ✅ Opt-in (no latency for images without forkd-agent)
- ❌ Brief wrong-clock window (~1-5 seconds between restore and `date -s`)
- ❌ Requires guest agent (forkd-agent.py must be installed in the image)
- ❌ Requires `date` binary (extremely unlikely to be absent, but handled)
- ❌ Adds ~1-2 seconds latency to restore when enabled (agent polling + exec)

### Option C: NTP/chrony in sandbox images
Install an NTP client in each image and let it sync on boot.

- ✅ Works on any platform
- ✅ Self-correcting (converges over time)
- ❌ Requires network egress (blocked by `none`/`restricted` network policy)
- ❌ Adds image dependencies (NTP package to every image)
- ❌ Takes 1-5 seconds to converge (not deterministic)
- ❌ Not reliable for CI sandboxes with restrictive policies

### Option D: forkd-agent.py host-time injection via kernel cmdline
Pass `forkd.clock_realtime=<ts>` as a kernel boot arg that forkd-agent.py reads and applies.

- ✅ No polling delay (clock set before any guest code runs)
- ❌ Boot args are snapshot-forward-only (same limitation as Option A)
- ❌ Could pass via agent TCP protocol instead, but that has the same polling window as Option B with more complexity

### Option E: KVM_CLOCK_REALTIME (Linux 5.16+)
Use `KVM_SET_CLOCK` with the `KVM_CLOCK_REALTIME` flag to restore host wall clock at the hypervisor level.

- ✅ Corrects clock before guest resumes (no wrong-clock window)
- ❌ x86-only
- ❌ Requires Linux ≥ 5.16 on the host
- ❌ Needs Firecracker support (may not exist)
- ❌ Firecracker-level change, not forkd-level

## Recommendation

**Options A and B are complementary**, not competing. They should both be merged:

- Option A (PR #289) narrows the wrong-clock window on x86 by keeping the guest clock host-synced at boot time.
- Option B (this PR) covers the gap on non-x86 architectures and for existing snapshots that haven't been re-baked. It also serves as a safety net: even with Option A deployed, if a snapshot was created before the fix or on a non-x86 host, Option B corrects the clock.

Option B is the **most portable** single solution and can work standalone. Option A is defense-in-depth.

## Changes

- **`crates/forkd-vmm/src/lib.rs`**:
  - Added `sync_guest_clocks()` (Phase 2.5 of `restore_many_with`): polls each child's guest agent, captures a fresh per-child host timestamp after ping, and execs `sh -c "date -s @<host_ts>"`.
  - Added `ClockSyncOutcome` enum (`Synced` / `Failed`) exposing per-child sync results, and `clock_sync_ms` + `clock_sync_outcomes` fields on `ForkResult`.
  - Added `ForkOpts::sync_guest_clocks` (default `false`) making clock sync opt-in.
  - Reject `n > 1` with `per_child_netns = false` (shared 10.42.0.2:8888 address makes per-child addressing impossible).
  - Added `GUEST_AGENT_ADDR` / `CLOCK_SYNC_AGENT_TIMEOUT` / `CLOCK_SYNC_EXEC_TIMEOUT` constants.
  - Unit tests: `clock_sync_date_command_format`, `fork_result_has_clock_sync_ms_field`, `clock_sync_outcome_construction_and_matching`, `fork_opts_sync_guest_clocks_defaults_disabled`, `restore_many_with_rejects_multi_child_shared_ip`.
  - Ignored integration tests (require Linux + KVM + firecracker + rootfs): `sync_guest_clocks_corrects_wrong_clock` (mock-clock isolation), `sync_guest_clocks_delayed_agent_readiness` (fresh-timestamp freshness).
- **`crates/forkd-controller/src/http.rs`**: set `sync_guest_clocks: true` on daemon spawn paths (`create_sandbox`, `spawn_one_for_workspace`) — all forkd images ship forkd-agent. CLI paths keep it `false` since they can't know whether an image has an agent.

Closes #288.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,3 +22,4 @@ build/
 # not in this public repo.
 /notes/
 Cargo.lock
+.pi-subagents/
--- a/crates/forkd-cli/src/main.rs
+++ b/crates/forkd-cli/src/main.rs
@@ -2277,6 +2277,7 @@ fn run_cmd(
                 prewarm_scratch_dir: None,
                 memory_backend: forkd_vmm::MemoryBackend::File,
                 enable_diff_snapshots: false,
+                sync_guest_clocks: false,
             },
             &work_dir,
         )
@@ -3128,6 +3129,7 @@ fn fork_cmd(
                 },
                 // CLI fork doesn't outlive its invocation, no diff snapshots.
                 enable_diff_snapshots: false,
+                sync_guest_clocks: false,
             },
             &work_dir,
         )
--- a/crates/forkd-controller/src/http.rs
+++ b/crates/forkd-controller/src/http.rs
@@ -1280,6 +1280,7 @@ async fn create_sandbox(
             // snapshots (see docs/design/diff-snapshots.md). The cost is
             // ~1 bit per page; negligible.
             enable_diff_snapshots: true,
+            sync_guest_clocks: true,
         };
         // Per-snapshot-tag work_dir would clash if two batches of the same tag
         // ran concurrently (e.g. two branches of the same source). Mix the
@@ -2803,6 +2804,7 @@ fn spawn_one_for_workspace(
         prewarm_scratch_dir: None,
         memory_backend: forkd_vmm::MemoryBackend::File,
         enable_diff_snapshots: true,
+        sync_guest_clocks: true,
     };
     let work_dir =
         std::env::temp_dir().join(format!("forkd-workspace-{snapshot_tag}-o{netns_offset}"));
--- a/crates/forkd-controller/src/http.rs
+++ b/crates/forkd-controller/src/http.rs
@@ -39,6 +39,7 @@ use crate::api::{
     VersionResponse, WorkspaceInfo, WorkspaceStatus,
 };
 use crate::state::Registry;
+use forkd_vmm::ClockSyncOutcome;
 
 const API_VERSION: &str = "v1";
 const BUILD_VERSION: &str = env!("CARGO_PKG_VERSION");
@@ -1366,6 +1367,16 @@ async fn create_sandbox(
         Err(e) => return server_error(&format!("blocking task panicked: {e}")),
     };
     if prewarm_requested {
+        let clock_failures: Vec<&str> = fork_result
+            .clock_sync_outcomes
+            .iter()
+            .filter_map(|o| match o {
+                ClockSyncOutcome::Failed { child_index, error } => {
+                    Some(format!("child {child_index}: {error}"))
+                }
+                _ => None,
+            })
+            .collect();
         tracing::info!(
             tag = %tag,
             n = fork_result.children.len(),
@@ -1373,6 +1384,7 @@ async fn create_sandbox(
             restore_ms = fork_result.restore_ms as u64,
             prewarm_ms = fork_result.prewarm_ms as u64,
             clock_sync_ms = fork_result.clock_sync_ms as u64,
+            clock_sync_failures =? clock_failures,
             "sandbox created (prewarmed)"
         );
     }
--- a/crates/forkd-controller/src/http.rs
+++ b/crates/forkd-controller/src/http.rs
@@ -1367,7 +1367,7 @@ async fn create_sandbox(
         Err(e) => return server_error(&format!("blocking task panicked: {e}")),
     };
     if prewarm_requested {
-        let clock_failures: Vec<&str> = fork_result
+        let clock_failures: Vec<String> = fork_result
             .clock_sync_outcomes
             .iter()
             .filter_map(|o| match o {
--- a/crates/forkd-controller/src/http.rs
+++ b/crates/forkd-controller/src/http.rs
@@ -1371,6 +1371,7 @@ async fn create_sandbox(
             spawn_ms = fork_result.spawn_ms as u64,
             restore_ms = fork_result.restore_ms as u64,
             prewarm_ms = fork_result.prewarm_ms as u64,
+            clock_sync_ms = fork_result.clock_sync_ms as u64,
             "sandbox created (prewarmed)"
         );
     }
--- a/crates/forkd-vmm/src/lib.rs
+++ b/crates/forkd-vmm/src/lib.rs
@@ -958,75 +958,68 @@ fn sync_guest_clocks(children: &[Vm]) -> (u128, Vec<ClockSyncOutcome>) {
         handles.push(thread::spawn(move || -> Result<()> {
             let addr = GUEST_AGENT_ADDR.to_string();
 
-            // Poll the guest agent until it responds or the timeout
-            // expires. The agent (forkd-agent.py) starts as PID 1 and
-            // begins listening on port 8888 as soon as init runs. For
-            // snapshot-restored VMs, the guest is already past kernel
-            // boot — only the agent's TCP listener needs to come up.
-            let deadline = Instant::now() + CLOCK_SYNC_AGENT_TIMEOUT;
-            loop {
-                let ping_result = match &netns {
-                    Some(ns) => ping_in_netns(ns, addr.clone()),
-                    None => ping_at(&addr),
-                };
-                if ping_result.is_ok() {
-                    break;
-                }
-                if Instant::now() >= deadline {
-                    bail!(
-                        "guest agent at {addr} did not respond within {}s",
-                        CLOCK_SYNC_AGENT_TIMEOUT.as_secs()
-                    );
-                }
-                thread::sleep(Duration::from_millis(200));
-            }
-
-            // Capture the host timestamp AFTER the agent is confirmed
-            // alive, immediately before exec. The old approach captured
-            // it before polling — a child that became ready near the
-            // 10-second deadline was set ~10 seconds in the past.
-            // Capturing per-child after ping minimizes the gap between
-            // timestamp capture and the date -s exec.
-            let now = SystemTime::now()
-                .duration_since(SystemTime::UNIX_EPOCH)
-                .unwrap_or_default()
-                .as_secs();
-            let date_cmd = format!("date -s @{now}");
-
-            // Agent is alive — set the guest clock to host wall time.
-            // We invoke `sh -c "date -s @<timestamp>"` because the guest
-            // agent's exec handler passes args directly to
-            // `subprocess.run(args)` — a bare `["-c", cmd]` would try
-            // to execute a program named "-c" and fail. The `sh -c`
-            // form runs the date command through the shell.
+            // Readiness probe: poll the guest agent until it responds
+            // or the timeout expires. The agent (forkd-agent.py) starts
+            // as PID 1 and begins listening on port 8888 as soon as init
+            // runs. For snapshot-restored VMs, the guest is already past
+            // kernel boot — only the agent's TCP listener needs to come
+            // up.
             //
-            // `date -s @<timestamp>` is supported by GNU coreutils date
-            // (≥ 8.5) and busybox date (with FEATURE_DATE_COMPAT). All
-            // forkd images are built from Debian/Ubuntu Docker bases,
-            // so GNU date is guaranteed.
-            let exec_result = match &netns {
-                Some(ns) => exec_in_netns(
-                    ns,
-                    addr.clone(),
-                    vec!["sh".into(), "-c".into(), date_cmd],
-                    CLOCK_SYNC_EXEC_TIMEOUT,
-                ),
-                None => exec_at(
-                    &addr,
-                    vec!["sh".into(), "-c".into(), date_cmd],
-                    CLOCK_SYNC_EXEC_TIMEOUT,
-                ),
+            // The probe is a closure so the timestamp-placement logic in
+            // `sync_one_child_clock` can be unit-tested with a
+            // deterministic delayed-readiness probe (see
+            // `sync_one_child_clock_timestamp_fresh_after_delayed_ping`).
+            let ping = || match &netns {
+                Some(ns) => ping_in_netns(ns, addr.clone()).map(|_| ()),
+                None => ping_at(&addr).map(|_| ()),
             };
 
-            match exec_result {
-                Ok(resp) if resp.exit_code == 0 => Ok(()),
-                Ok(resp) => bail!(
-                    "date -s exited with {}: {}",
-                    resp.exit_code,
-                    resp.stderr.trim()
-                ),
-                Err(e) => Err(e),
-            }
+            // Timestamp source: host wall time. Captured per child,
+            // AFTER ping succeeds — see `sync_one_child_clock`.
+            let now = || {
+                SystemTime::now()
+                    .duration_since(SystemTime::UNIX_EPOCH)
+                    .unwrap_or_default()
+                    .as_secs()
+            };
+
+            // Clock-setter: invoke `sh -c "date -s @<ts>"` via the guest
+            // agent's exec. We use `sh -c` because the guest agent's exec
+            // handler passes args directly to `subprocess.run(args)` — a
+            // bare `["-c", cmd]` would try to execute a program named
+            // "-c" and fail.
+            //
+            // `date -s @<ts>` is supported by GNU coreutils date (≥ 8.5)
+            // and busybox date (with FEATURE_DATE_COMPAT). All forkd
+            // images are built from Debian/Ubuntu Docker bases, so GNU
+            // date is guaranteed.
+            let set_clock = |ts: u64| {
+                let date_cmd = format!("date -s @{ts}");
+                let exec_result = match &netns {
+                    Some(ns) => exec_in_netns(
+                        ns,
+                        addr.clone(),
+                        vec!["sh".into(), "-c".into(), date_cmd],
+                        CLOCK_SYNC_EXEC_TIMEOUT,
+                    ),
+                    None => exec_at(
+                        &addr,
+                        vec!["sh".into(), "-c".into(), date_cmd],
+                        CLOCK_SYNC_EXEC_TIMEOUT,
+                    ),
+                };
+                match exec_result {
+                    Ok(resp) if resp.exit_code == 0 => Ok(()),
+                    Ok(resp) => bail!(
+                        "date -s exited with {}: {}",
+                        resp.exit_code,
+                        resp.stderr.trim()
+                    ),
+                    Err(e) => Err(e),
+                }
+            };
+
+            sync_one_child_clock(ping, now, set_clock)
         }));
     }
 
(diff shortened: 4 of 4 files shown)


jrimmer:
## Update: integration test + exec args bug fix

### Bug fix in sync_guest_clocks

The exec args were `["-c", cmd]` which caused the guest agent's `subprocess.run(["-c", "date -s @..."])` to fail with `FileNotFoundError` — it tried to execute a program named `-c`. The clock sync was silently failing every time (best-effort, logged as warning). This was masked because the `clocksource=kvm-clock` boot arg (#289) was correcting the clock on modern kernels via `KVM_CLOCK_REALTIME`.

**Fixed:** changed to `["sh", "-c", cmd]` so the date command runs through the shell as intended.

### Integration test: `sync_guest_clocks_corrects_wrong_clock`

Added `#[ignore]` integration test that isolates the scripting solution:

1. Boots a fresh VM
2. Deliberately sets the guest clock to 2001-09-09 (`date -s @1000000000`)
3. Verifies the guest clock IS wrong
4. Calls `sync_guest_clocks` on the VM
5. Verifies the guest clock is corrected to host wall time (within 5s)
6. Verifies the corrected time is NOT the old mock time

This tests the *mechanism* (date -s via guest agent exec), not the *environment* (kernel clock behavior). Even if `KVM_CLOCK_REALTIME` already corrected the clock, the test forces a wrong clock and verifies the sync function fixes it.

### Complementarity with #289

#289's test (`kvm_clock_survives_snapshot_restore`) verifies the boot arg works on the restore path for modern x86 kernels (≥5.16 with `KVM_CLOCK_REALTIME`). #300's test verifies the scripting solution works regardless of kernel version or architecture. Together they cover all scenarios:

| Scenario | #289 test | #300 test |
|---|---|---|
| Modern x86 restore | ✅ verifies kvm-clock auto-corrects | ✅ but uses mock clock |
| Older x86 restore | ❌ expected fail (no KVM_CLOCK_REALTIME) | ✅ mock clock isolates the fix |
| aarch64 restore | ❌ kvm-clock is x86-only | ✅ mock clock isolates the fix |

WaylandYang (changes requested):
The corrected sh invocation proves the basic date mechanism, but the restore integration still has blocking correctness/API issues:

1. The host timestamp is captured before polling any agent. A child that becomes ready near the 10-second deadline is deliberately set up to roughly 10 seconds in the past. Capture SystemTime::now() per child after ping succeeds, immediately before exec (ideally preserve sub-second precision or account for round-trip delay).
2. restore_many_with unconditionally performs this agent protocol. Custom/no-agent snapshots now add up to 10 seconds to every restore even though failure is only a warning. Make clock sync an explicit capability/option with a bounded configurable timeout, or limit it to snapshots known to contain forkd-agent.
3. For n>1 with per_child_netns=false, every worker contacts the same 10.42.0.2:8888. That does not identify a different child; requests can all hit one VM while siblings remain unsynchronized (and the shared IP/MAC topology is conflicting in the first place). #281's controller validation does not cover the CLI or public Snapshot::restore_many_with API, and this PR is based on main. Reject/disable this mode here or provide unique addressing.
4. A best-effort warning still returns a successful sandbox with the TLS-breaking stale clock. Please expose per-child sync outcomes so callers can choose fail-closed/degraded behavior, and align the claim to the actual guarantee.

Add tests for delayed agent readiness (timestamp freshness), no-agent latency, and multi-child address selection—not only a single already-running VM.

jrimmer:
## Fix: per-child timestamp, opt-in clock sync, reject n>1 shared-IP

**Finding 1 (timestamp before polling):** The host timestamp was captured before polling the guest agent — a child that became ready near the 10-second deadline was set ~10 seconds in the past.

**Fix:** `SystemTime::now()` is now captured per child **after** ping succeeds, immediately before the `date -s` exec. This minimizes the gap between timestamp capture and clock correction.

**Finding 2 (unconditional agent protocol):** `restore_many_with` unconditionally performed the agent protocol — custom/no-agent snapshots added up to 10 seconds to every restore even though failure is only a warning.

**Fix:** Clock sync is now opt-in via `ForkOpts::sync_guest_clocks` (default `false`). The daemon's spawn paths (`create_sandbox`, `spawn_one_for_workspace`) set it to `true` because all forkd images contain forkd-agent. CLI paths keep it `false` — the CLI doesn't know whether the image has an agent. Custom images that don't include forkd-agent no longer pay the 10-second timeout on every restore.

**Finding 3 (n>1 shared-IP addressing):** For n>1 with `per_child_netns=false`, every worker contacted the same `10.42.0.2:8888` — the shared IP/MAC topology is fundamentally conflicting (requests can all hit one VM while siblings remain unsynchronized).

**Fix:** `restore_many_with` now rejects `n > 1` with `per_child_netns=false` with a clear error message directing callers to use `per_child_netns=true` for multi-child spawns.

jrimmer:
## Update: per-child clock sync outcomes exposed

The clock sync function now returns per-child outcomes via `ForkResult::clock_sync_outcomes: Vec<ClockSyncOutcome>`, where `ClockSyncOutcome` is an enum with `Synced { child_index }` and `Failed { child_index, error }` variants. Callers can inspect this to fail-closed or degrade when some children have unsynchronized clocks.

The controller's `create_sandbox` now logs `clock_sync_failures` alongside the existing timing fields, listing the child index and error for each failure.

This addresses the finding that a best-effort warning still returns a successful sandbox with a potentially stale clock — the per-child outcomes are now visible to the caller for fail-closed/degraded decisions.

Tests:
- `clock_sync_outcome_construction_and_matching`: verifies enum discriminants, field access, and the filter-for-failures pattern callers use.
- The `#[ignore]` integration test (`sync_guest_clocks_corrects_wrong_clock`) now also asserts the outcome vec contains exactly one `Synced` entry for the single child.

WaylandYang (changes requested):
Thank you for addressing the main API/design feedback: timestamp capture is now after ping, clock sync is opt-in, shared-IP multi-child restore is rejected, and per-child outcomes are exposed. The current head is not mergeable yet for two concrete reasons.

Rust CI is failing on Clippy because the iterator enumerates children and then discards the index. Please fix the warning and restore a fully green required check. The requested behavioral coverage is also still missing: delayed agent readiness (to prove timestamp freshness), a no-agent/disabled-sync latency bound, and multi-child address/mode behavior. The enum-construction test only verifies data shape and does not exercise those restore paths.

Please add those regression tests and update the PR description, which still describes capturing one timestamp before polling even though the implementation changed. Thanks for the thoughtful iteration; I’ll re-review once CI and the end-to-end edge cases are covered.

jrimmer:
Addressed the CI failure and the missing behavioral coverage:

- Fixed the clippy `unused_enumerate_index` warning — the `sync_guest_clocks` child loop no longer calls `.enumerate()` and discards the index.
- Added a unit test for the `n>1` + `per_child_netns=false` shared-IP rejection in `restore_many_with`.
- Added a unit test pinning the opt-in default (`sync_guest_clocks` is `false` by default, so no-agent images get no added latency).
- Added a delayed-agent-readiness integration test that boots a VM, calls `sync_guest_clocks` immediately (no pre-wait, so it must poll for the not-yet-ready agent), and asserts the guest clock matches host wall time within 3s.
- Fixed a `Vec<&str>`/`String` type mismatch and a `useless_vec` lint in the test code.

Updated the PR description to reflect that the timestamp is captured per-child after the agent responds (not once before polling).

Verified: `cargo clippy` clean and the vmm test suite passes (46 tests, 2 ignored integration tests).

WaylandYang (changes requested):
Thank you for fixing Clippy, adding the mode/default coverage, and updating the description. The implementation now captures the timestamp after ping, keeps sync opt-in, rejects shared-IP multi-child restore, and exposes per-child outcomes.

The branch currently conflicts with main and has no CI for this head, so please rebase and resolve the overlap with merged #289 first. The delayed-readiness test also does not deterministically create a delay: it calls sync immediately, but if the agent is ready within the 3-second tolerance, the old pre-poll timestamp bug would still pass. Please introduce an injectable/test readiness probe or an explicit delayed-agent fixture, assert that the poll delay exceeds the accepted clock drift, and then verify the final timestamp remains fresh.

Finally, sign off every resulting commit with the README-required DCO trailer (`git commit -s`); the latest commit currently has none. After the rebase, deterministic regression coverage, sign-off, and green CI, this should be ready for final review.

WaylandYang (changes requested):
Thank you for rebasing, restoring green CI, and signing the commits. The implementation changes requested earlier are present, but the delayed-readiness regression remains non-deterministic. sync_guest_clocks_delayed_agent_readiness calls sync immediately and only checks the final drift plus an upper elapsed bound; it never guarantees that agent readiness was delayed beyond the accepted 3-second drift. If the agent is ready quickly, the old pre-poll timestamp implementation still passes this test.

Please inject a readiness probe/clock or use an explicitly delayed agent fixture, assert that the induced poll delay exceeds the accepted drift, and then verify that the final guest timestamp is fresh. This is the remaining blocker; the mode validation, opt-in behavior, timestamp placement, and per-child outcomes otherwise look good.

WaylandYang:

jrimmer:
## Update: deterministic delayed-agent-readiness test (round 6)

Thank you for the final round. The delayed-readiness regression was non-deterministic — it called sync immediately and only checked the final drift plus an upper elapsed bound, so it never guaranteed that agent readiness was actually delayed beyond the accepted 3-second drift. If the agent came up quickly, the old pre-poll-timestamp implementation still passed.

### Fix

The timestamp-placement logic is extracted from `sync_guest_clocks` into `sync_one_child_clock(ping, now, set_clock, timeout)`, which accepts **injectable** readiness-probe / timestamp-source / clock-setter / timeout closures. The production call site wires these to `ping_at`/`ping_in_netns`, `SystemTime::now`, the `date -s` exec, and `CLOCK_SYNC_AGENT_TIMEOUT`. The test wires them to a deterministic probe that **fails for 4 seconds** (> the 3s accepted drift) then succeeds.

### The deterministic test: `sync_one_child_clock_timestamp_fresh_after_delayed_ping`

It runs in the unit-test suite on every platform — **no VM, guest agent, or KVM required**. It asserts:

1. **The induced poll delay (4s) exceeds the accepted drift (3s)** — this is what makes the test meaningful. A compile-time `const _: () = { assert!(INDUCED_DELAY_SECS > ACCEPTED_DRIFT_SECS) }` guarantees the invariant can never silently invert. If the probe became ready before the accepted drift, the freshness assertion could not distinguish fresh from stale.
2. **The timestamp was captured AFTER ping succeeded** (`now_called_at >= probe_ready_at`, freshness_gap < 3s). A pre-poll-timestamp implementation would fail this by ~4s — `now_called_at` would be set before the loop, ~4s before `probe_ready_at`.
3. **`set_clock` was actually invoked with that fresh timestamp** (`recorded_ts == now_at`).

### Additional coverage from ce-code-review

An independent 5-reviewer code review (correctness, testing, maintainability, reliability, adversarial) flagged two more coverage gaps, now addressed:

- **`sync_one_child_clock_times_out_when_probe_never_succeeds`** — uses a 500ms injected timeout (the timeout is now an injected parameter, so the test doesn't wait the full 10s) to verify the bail message and that `now()`/`set_clock` are never called on the timeout path.
- **`sync_one_child_clock_propagates_set_clock_failure`** — verifies a `set_clock` error propagates verbatim and that `now()` was still called (timestamp captured before the failed set).
- **Timeout error now includes the agent address** — the refactor had dropped `{addr}` from the timeout message; it's now restored at the call site via `.map_err(|e| e.context(format!("guest agent at {addr} did not come up")))` so operators can identify which child failed in multi-child restores.

### Verification

- `cargo clippy` clean for forkd-vmm/controller/cli/uffd (`--all-targets --all-features -- -D warnings`).
- `cargo test -p forkd-vmm --lib`: **50 passed, 0 failed, 3 ignored** (the 3 ignored are the KVM-requiring integration tests).
- All commits signed off with the DCO `Signed-off-by` trailer.

The end-to-end mechanism (real agent + `date -s`) is still covered by the `#[ignore]` integration test `sync_guest_clocks_corrects_wrong_clock`.

jrimmer:
Thanks for the careful re-review. The delayed-readiness regression you flagged is now deterministic — the non-deterministic `sync_guest_clocks_delayed_agent_readiness` integration test was replaced in `be339d4` (pushed after your 18:28 round) with a unit test that injects the probe directly.

**New test: `sync_one_child_clock_timestamp_fresh_after_delayed_ping`** (`crates/forkd-vmm/src/lib.rs`) exercises the extracted `sync_one_child_clock(ping, now, set_clock, timeout)` core via injected closures — no VM, guest agent, or KVM required, runs in the unit-test suite on every platform:

1. **Induced delay > accepted drift.** The readiness probe deterministically fails for 4 s (> the 3 s accepted drift) then succeeds. The test asserts `poll_elapsed >= 4 s`, so a probe that became ready early would fail the test rather than silently passing.
2. **Timestamp captured after ping.** `now()` records its wall-time call; the probe records when it first returns `Ok`. The test asserts `now_at >= ready_at` — a pre-poll-timestamp implementation would capture `now()` ~4 s before `ready_at` and fail this assertion by ~4 s.
3. **Timestamp is fresh.** `freshness_gap = now_at - ready_at` is asserted `<= 3 s` (the accepted drift). The old pre-poll implementation would produce a ~4 s gap and fail here.
4. **`set_clock` invoked with the fresh timestamp.** Asserts the recorded `set_clock` ts equals the `now()` return, so the timestamp actually sent to the guest is the post-ping one.

The end-to-end mechanism (real agent + `date -s`) is still covered by the `#[ignore]` integration test `sync_guest_clocks_corrects_wrong_clock`. All commits are DCO-signed, the branch is rebased on current main, and `cargo clippy -p forkd-vmm --lib --tests --target x86_64-unknown-linux-gnu -- -D warnings` is clean. Could you re-review the head and let me know if the deterministic coverage meets the bar?

WaylandYang (approved):
The current head addresses the remaining blocker. I verified that the delayed-readiness test deterministically holds the probe for 4 seconds (beyond the 3-second freshness tolerance), captures the timestamp only after readiness, and proves that exact fresh value reaches the clock setter. The opt-in/default behavior, shared-IP multi-child rejection, per-child outcomes, timeout/failure paths, DCO, and required CI are also in order.

Approved. Non-blocking follow-up: the context `guest agent ... did not come up` is also attached to a clock-set failure; a more precise message would improve diagnostics, but this does not affect correctness.

Issue #301, opened 2026-08-13
Controller restore-path resilience: stop mass-pruning sandboxes on restart and restore failures


jrimmer:
## Problem

Intermittent CI failures across the stack trace back to a multi-layer reliability gap in the forkd controller's snapshot restore path. The failure chain:

1. **`restore_many_with` fails** — firecracker VMs fail to start during restore, logging `socket /tmp/forkd-daemon-<tag>-<id>/child-N.sock never appeared within 10s` (forkd-vmm `wait_for_sock`, lib.rs ~line 902). Observed to be systemic (affects multiple image tags, not one).
2. **Controller restart prunes ALL sandboxes** — on startup, `Registry::reconcile()` (state.rs:213) prunes every sandbox entry whose PID is no longer alive. After a watchdog-triggered restart, *every* running sandbox is pruned (observed: 22 pruned in one incident), invalidating all upstream lease references in a single stroke.
3. **Orphaned firecracker processes survive backend restarts** — when the upstream lease manager restarts, its `Shutdown()` path does not fully kill all sandbox VMs via the forkd API, leaving orphaned firecracker processes. These orphans cause resource contention (tap device conflicts, netns conflicts, fd limits) that prevents new `restore_many_with` operations from succeeding, producing a cascade.

### Root cause analysis

- The 10-second `wait_for_sock` timeout (`Duration::from_secs(10)` at lib.rs:980 and :1486) is a fixed budget with no load-awareness. Under contention from orphaned VMs or parallel restore bursts, firecracker processes take longer than 10s to open their API socket, and the wait bails out.
- `reconcile()` is a blunt instrument: it cannot distinguish "VM died unexpectedly" from "VM is alive but the controller just restarted." On restart, *all* live VMs become unreconcilable (no `live_vms` handles can be reconstructed from registry data alone) and are pruned.
- There is no pre-restore orphan detection: `restore_many_with` does not check whether firecracker processes from a prior (crashed) run still hold tap devices or netns slots before attempting a new restore.

### Resource exhaustion ruled out

The host has ample capacity during failures: 62 GB RAM (58 GB available), 22 GB disk free, 32 GB tmpfs. The 22 firecracker processes and 4m37s CPU / 2.7 GB peak of the controller are well within limits. The failures are not resource exhaustion — they are startup-path contention and blunt-restart side effects.

### 1. Restore-path resilience
[...]

PR #302, opened 2026-08-13
fix(vmm): restore-path resilience — configurable timeout, per-VM partial-failure reporting, pre-restore orphan detection

jrimmer:
## Summary

Partially addresses #301. Three changes to stop a single child failure from dooming the whole restore batch, and to surface contention from orphaned firecracker processes before a restore is attempted.

**Issue #301 remains open** — the orphan-reap/gate and graceful-reconcile criteria are not yet satisfied. This PR narrows to the restore-path resilience subset (criteria #2 and #4) and documents the remaining gaps.

## Problem

Intermittent failures trace back to the forkd controller's snapshot restore path:
1. The `wait_for_sock` timeout was hardcoded at 10s — under load it's too tight.
2. `restore_many_with` bailed on the first child failure, silently dropping already-spawned siblings.
3. Orphaned firecracker processes from a prior crash hold tap/netns resources and doom new restores, but were invisible until a restore failed.

### 1. Configurable per-child socket wait timeout (`ForkOpts` field)
- New `ForkOpts::socket_wait_timeout_secs` (default 10s, matching the historical budget) drives the `wait_for_sock` calls in `restore_many_with`.
- Daemon paths (`create_sandbox`, `spawn_one_for_workspace`) default to 30s, configurable via `FORKD_SOCKET_WAIT_TIMEOUT_SECS` env var.
- A warning fires when a child takes >80% of the budget to appear, so operators see contention before it pushes the next spawn over.
- `wait_for_sock` now bails when the socket path is a directory (not a Unix socket) — previously `Path::exists()` returned true for a directory, masking the failure.
- CLI paths keep the 10s default.

### 2. Per-VM partial-failure reporting (`RestoreError`/`RestoreFailure`/`ForkChild`)
- New types: `RestorePhase` (enum), `RestoreFailure` (struct), `RestoreError` (struct with `failures: Vec<RestoreFailure>`), `ForkChild` (struct with `child_index` + `vm`), implementing `Display + std::error::Error` where applicable.
- `ForkResult.children` is now `Vec<ForkChild>` — each surviving child carries its 1-based within-batch `child_index` so callers can map a surviving VM back to its resource slot (netns/cgroup at `netns_offset + child_index`). This preserves the child-to-VM mapping that a plain `Vec<Vm>` would lose when failures are filtered out.
- `restore_many_with` now collects per-child results in each phase instead of bailing on the first failure:
  - **Spawn phase**: failures recorded but remaining children still spawn; on any failure, spawned children are dropped (killing firecracker) and a `RestoreError` carrying every failure is returned.
  - **Socket-wait phase**: a timed-out child is recorded; on any timeout, all children are dropped and a `RestoreError` is returned. Concurrent socket wait: one thread per child, all with the SAME timeout so the batch is bounded by one deadline, not n × timeout.
  - **Restore phase**: each thread's result is collected; a panicked thread is recorded as a restore failure rather than panicking the caller.
- `RestoreError` names every failed child (index, phase, error, pid when known) so callers can report which specific child failed rather than a single bail that loses batch context.

### 3. Pre-restore orphan detection (`scan_firecracker_orphans`)
- New `OrphanProcess` struct + `scan_firecracker_orphans(known_pids)` function scans `/proc` for running firecracker processes whose PIDs are NOT in the caller's `live_vms` set.
- Linux impl: reads `/proc/<pid>/comm` (exact "firecracker" match), `/proc/<pid>/cmdline`, `/proc/<pid>/stat` (elapsed time).
- Non-Linux stub: returns empty vec (for dev-box compilation).
- Controller calls it before both `create_sandbox` and `spawn_one_for_workspace` restores via a `warn_orphan_firecrackers` helper that scopes the `live_vms` lock to just PID collection, dropping it before the `/proc` scan so the scan doesn't serialize concurrent sandbox operations.
- Exposed as `forkd_orphan_firecrackers_detected_total` metric so operators can alert on the precursor.
- Read-only — never kills anything. The scan is a point-in-time observation with a documented TOCTOU window.

## Controller-level all-or-nothing semantics

The controller's `create_sandbox` handler is intentionally **all-or-nothing**: a sandbox requires all N children, so when `restore_many_with` returns partial success (some children restored, some failed), the controller drops the surviving children (kills their firecracker processes) and either retries the whole batch (if the failure is a transient "busy" condition) or returns a structured error. The per-child failure reporting from `restore_many_with` is used at the controller level for **diagnostics and retry classification only** — partial success is NOT exposed through the API.

This is a deliberate design choice, not a limitation to be fixed later: a half-populated sandbox is worse than no sandbox (it would hold resources but be unable to serve requests). Issue #301 tracks the broader partial-success-through-controller discussion; this PR does not close it.

## Known limitations (issue #301 remains open)

1. **Orphan detection is diagnostic-only**: the pre-restore scan logs and counts orphans but does not reap or gate on them. A lifecycle-level orphan reap/gate path (kill or block on untracked firecracker processes before restore) is tracked in #301.

2. **Graceful reconcile is tracked separately**: the controller restart mass-prune issue (#301 criterion #5) is addressed by #299 (startup orphan kill) and #304 (pidfd durable identity), not this PR.

3. **Controller restart does not reattach surviving VMs**: #301 criterion #5 (a controller restart should not mass-prune alive VMs) requires reconcile() changes that distinguish "VM died" from "controller crashed with surviving VMs". This is out of scope for this PR.

## Tests
- `restore_error_display_lists_all_failures` — Display lists every failure with phase, pid, and error.
- `restore_error_single_failure_display` — single Spawn-phase failure (no pid).
- `restore_phase_eq_distinguishes_phases` — phase enum equality.
- `scan_firecracker_orphans_non_linux_returns_empty` — stub contract on non-Linux.
- `scan_firecracker_orphans_empty_known_set_non_linux` — empty known set contract.
- `orphan_process_struct_shape` — OrphanProcess field stability.
- `restore_many_partial_spawn_failure_keeps_siblings` (ignored, requires Linux + KVM + root) — integration test: boots a parent VM, snapshots it, restores two children with per-child netns. Child 1's netns is pre-provisioned; child 2's is not, so `ip netns exec` fails at spawn. Asserts child 1 survives with `child_index=1`, child 2 fails at Spawn phase.

## Out of scope
The concurrent-RW-mount-of-shared-rootfs issue (`restore_many_with` sends the same vmstate to all children) needs an immutable baseline + per-VM writable layer and is tracked as a separate architectural initiative. The prewarm (Phase 3) and memfd (Phase 1.5) paths retain their pre-existing early-bail behavior — they are host-resource failures rather than per-VM failures, and converting them is outside this change's scope.

### DIFF
--- a/crates/forkd-cli/src/main.rs
+++ b/crates/forkd-cli/src/main.rs
@@ -2278,10 +2278,19 @@ fn run_cmd(
                 memory_backend: forkd_vmm::MemoryBackend::File,
                 enable_diff_snapshots: false,
                 sync_guest_clocks: false,
+                socket_wait_timeout_secs: forkd_vmm::DEFAULT_SOCKET_WAIT_SECS,
             },
             &work_dir,
         )
         .context("restore_many failed")?;
+    if result.children.is_empty() {
+        let detail = result
+            .failures
+            .first()
+            .map(|f| format!("child {} {:?}: {}", f.child_index, f.phase, f.error))
+            .unwrap_or_else(|| "no children restored".to_string());
+        bail!("restore_many produced no live child: {detail}");
+    }
 
     // Wait for the agent to come up, then exec the command.
     let target = "10.42.0.2:8888".to_string();
@@ -2307,8 +2316,8 @@ fn run_cmd(
     let exit = resp.exit_code;
 
     // Shutdown.
-    for c in &result.children {
-        let _ = c.shutdown();
+    for fc in &result.children {
+        let _ = fc.vm.shutdown();
     }
     drop(result);
 
@@ -3130,15 +3139,25 @@ fn fork_cmd(
                 // CLI fork doesn't outlive its invocation, no diff snapshots.
                 enable_diff_snapshots: false,
                 sync_guest_clocks: false,
+                socket_wait_timeout_secs: forkd_vmm::DEFAULT_SOCKET_WAIT_SECS,
             },
             &work_dir,
         )
         .context("restore_many_with failed")?;
+    if !result.failures.is_empty() {
+        for f in &result.failures {
+            eprintln!(
+                "⚠ child {} {:?} failed: {}",
+                f.child_index, f.phase, f.error
+            );
+        }
+    }
 
     let total_ms = result.spawn_ms + result.restore_ms;
     println!("✓ all sockets up in {} ms", result.spawn_ms);
     println!(
-        "✓ {n} restores fired in parallel in {} ms",
+        "✓ {} / {n} restores fired in parallel in {} ms",
+        result.children.len(),
         result.restore_ms
     );
     println!("✓ total wall-clock: {total_ms} ms");
@@ -3146,12 +3165,12 @@ fn fork_cmd(
     eprintln!("==> letting children settle for {settle_secs}s...");
     thread::sleep(Duration::from_secs(settle_secs));
 
-    let alive = result.children.iter().filter(|c| c.is_alive()).count();
+    let alive = result.children.iter().filter(|fc| fc.vm.is_alive()).count();
     println!("✓ {alive} / {n} children alive");
 
     eprintln!("==> shutting down...");
-    for c in &result.children {
-        let _ = c.shutdown();
+    for fc in &result.children {
+        let _ = fc.vm.shutdown();
     }
     thread::sleep(Duration::from_secs(2));
     drop(result); // triggers kill via Drop for any still alive
--- a/crates/forkd-controller/src/http.rs
+++ b/crates/forkd-controller/src/http.rs
@@ -2958,7 +2958,6 @@ fn spawn_one_for_workspace(
         enable_diff_snapshots: true,
         sync_guest_clocks: true,
         socket_wait_timeout_secs: daemon_socket_wait_timeout_secs(),
-
     };
     let work_dir =
         std::env::temp_dir().join(format!("forkd-workspace-{snapshot_tag}-o{netns_offset}"));
--- a/crates/forkd-controller/src/http.rs
+++ b/crates/forkd-controller/src/http.rs
@@ -92,6 +92,11 @@ pub struct AppState {
     /// sandbox and workspace spawn paths so concurrent spawns never
     /// pick overlapping offsets (security review #282).
     pub netns_alloc: std::sync::Arc<crate::netns::NetnsAllocator>,
+    /// Monotonic count of pre-restore orphan scans that found firecracker
+    /// processes not tracked by `live_vms`. Exposed via `/metrics` as
+    /// `forkd_orphan_firecrackers_detected_total` so operators can alert
+    /// on the tap/netns/fd-collision precursor instead of relying on logs.
+    pub orphan_firecrackers_detected: std::sync::atomic::AtomicU64,
     /// Scratch directory used for prewarm throwaway snapshots when
     /// `CreateSandboxRequest::prewarm` is set. Mirror of
     /// `DaemonConfig::prewarm_scratch_dir`.
@@ -243,6 +248,9 @@ async fn metrics(State(s): State<SharedState>) -> impl IntoResponse {
     let (snap_count, sb_count) = s.registry.counts();
     let branches_in_flight = s.branch_in_flight.lock().len();
     let branch_cap = s.branch_concurrency_cap;
+    let orphans = s
+        .orphan_firecrackers_detected
+        .load(std::sync::atomic::Ordering::Relaxed);
     // Prometheus text format. Keep names stable — exporters depend on them.
     let body = format!(
         "# HELP forkd_snapshots_total Number of snapshots known to the controller.\n\
@@ -257,6 +265,9 @@ async fn metrics(State(s): State<SharedState>) -> impl IntoResponse {
          # HELP forkd_branch_concurrency_cap Configured maximum concurrent BRANCH operations.\n\
          # TYPE forkd_branch_concurrency_cap gauge\n\
          forkd_branch_concurrency_cap {branch_cap}\n\
+         # HELP forkd_orphan_firecrackers_detected_total Cumulative firecracker orphans detected.\n\
+         # TYPE forkd_orphan_firecrackers_detected_total counter\n\
+         forkd_orphan_firecrackers_detected_total {orphans}\n\
          # HELP forkd_build_info Build version of the controller binary.\n\
          # TYPE forkd_build_info gauge\n\
          forkd_build_info{{version=\"{BUILD_VERSION}\"}} 1\n"
@@ -312,6 +323,44 @@ async fn list_snapshots(State(s): State<SharedState>) -> impl IntoResponse {
         }
     }
     Json(snapshots)
+}
+
+/// #301: scan for orphaned firecracker processes before a restore. An
+/// orphan is a firecracker PID on the host that is NOT tracked by
+/// `live_vms` — leftovers from a prior (crashed) run that may hold tap
+/// devices, netns slots, or fd budget and doom a new restore to
+/// socket/restore failures. Best-effort, read-only: it never kills.
+/// The lock is held only long enough to snapshot the known-PID set;
+/// the `/proc` scan + logging run after the lock is released so the
+/// scan doesn't serialize concurrent sandbox operations.
+fn warn_orphan_firecrackers(s: &SharedState, tag: &str) {
+    // Collect known PIDs under a short-lived guard, then drop the lock
+    // before the blocking `/proc` readdir + per-PID reads.
+    let known: std::collections::HashSet<u32> = {
+        let live = s.live_vms.lock();
+        live.values().map(|vm| vm.pid()).collect()
+    };
+    let orphans = forkd_vmm::scan_firecracker_orphans(&known);
+    if orphans.is_empty() {
+        return;
+    }
+    s.orphan_firecrackers_detected
+        .fetch_add(orphans.len() as u64, std::sync::atomic::Ordering::Relaxed);
+    tracing::warn!(
+        tag = %tag,
+        count = orphans.len(),
+        pids = ?orphans.iter().map(|o| o.pid).collect::<Vec<_>>(),
+        "orphaned firecracker processes detected before restore; \
+         these may hold tap/netns resources and cause spawn failures"
+    );
+    for o in &orphans {
+        tracing::warn!(
+            pid = o.pid,
+            elapsed_secs = o.elapsed_secs,
+            cmdline = %o.cmdline,
+            "orphaned firecracker process"
+        );
+    }
 }
 
 /// Phase 6.4: lazy reaper. Walks `AppState::live_in_flight`, joins
@@ -1224,6 +1273,15 @@ async fn create_sandbox(
     let prewarm_requested = req.prewarm;
     let mut snapshot = snapshot;
     let head_tag_for_log = req.snapshot_tag.clone();
+    // #301: surface orphaned firecracker processes before the restore.
+    // The helper scopes the live_vms lock to just PID collection and
+    // runs the /proc scan + logging after the lock is released.
+    //
+    // Limitation: this scan is point-in-time and does NOT guarantee that
+    // no orphan appears between the scan and the restore. It is a
+    // diagnostic aid, not a gate. Issue #301 remains open for the
+    // lifecycle-level orphan reap/gate work.
+    warn_orphan_firecrackers(&s, &head_tag_for_log);
     let tap_token_for_closure = tap_token.clone();
     let (fork_result, mut netns_reservation, mut tap_claim) = match tokio::task::spawn_blocking(move || -> anyhow::Result<(forkd_vmm::ForkResult, Option<crate::netns::NetnsReservation>, Option<SharedTapClaim>)> {
         // Reserve the netns offset range INSIDE the blocking task so
@@ -1282,6 +1340,12 @@ async fn create_sandbox(
             // ~1 bit per page; negligible.
             enable_diff_snapshots: true,
             sync_guest_clocks: true,
+            // Under spawn bursts (warm-pool refill) the default 10s socket
+            // wait can be too tight when prior-VM tap/netns teardown races
+            // with new firecracker spawns. Give the daemon path a more
+            // generous budget so a single slow child doesn't fail the batch.
+            socket_wait_timeout_secs: daemon_socket_wait_timeout_secs(),
+
         };
         // Per-snapshot-tag work_dir would clash if two batches of the same tag
         // ran concurrently (e.g. two branches of the same source). Mix the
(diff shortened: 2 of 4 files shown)


jrimmer:
## Fix summary

Follow-up commit `e3a3e5c` addresses findings raised on this PR:

**Correctness / dead code:**
- Removed the dead `sock_ok` vec in the socket-wait loop — it was collected on the success path but only discarded via `let _ = sock_ok`. The success path proceeds with all children regardless.
- Removed the dead `now_secs` variable in `scan_firecracker_orphans_impl` — it was computed once at the top of the scan but never read; the per-orphan elapsed-time computation calls `boot_relative_secs()` itself.

**Misleading comments:**
- `proc_clk_tck` was a function that unconditionally returned `Some(100)`; the `unwrap_or(100)` at the call site and the `hz == 0` guard were unreachable. Replaced with a `PROC_CLK_TCK` const. Fixed the doc comment that incorrectly claimed `libc` would add a Linux-only dep — `libc` is already a `forkd-vmm` dependency.
- `scan_firecracker_orphans` doc claimed the scan was "one `/proc` readdir + per-PID `/proc/<pid>/comm` read" but the implementation also reads `/proc/<pid>/cmdline` and `/proc/<pid>/stat` per orphan. Corrected the doc to reflect the actual per-orphan I/O, and added a TOCTOU note: the scan is a point-in-time observation, and PIDs may be reused between scan and restore (the scan is logging-only and never gates a restore).
- Removed a dead cross-reference to a nonexistent `kill_orphans` path in `state.rs`.

**Lock scoping / contention:**
- The orphan scan previously held the `live_vms` Mutex across the entire `/proc` readdir + per-PID file reads + logging, serializing concurrent sandbox operations. Extracted a `warn_orphan_firecrackers` helper that scopes the lock to just PID collection, dropping it before the scan runs. This also deduplicates the scan between `create_sandbox` and `spawn_one_for_workspace`.

**Env configurability:**
- The daemon socket-wait timeout (30s) is now configurable via the `FORKD_SOCKET_WAIT_TIMEOUT_SECS` env var, so operators can tune without a redeploy.

The prewarm (Phase 3) and memfd (Phase 1.5) paths retain their pre-existing early-bail behavior — they are host-resource failures rather than per-VM failures, and converting them is outside this PR's scope.

WaylandYang (changes requested):
Thank you for improving error reporting and making the socket wait configurable; the structured per-child failures and lock scoping are useful changes. The implementation still does not provide the resilience claimed by this PR or fully close #301.

The socket waits are performed serially with a full timeout per child, so total delay can grow to `n × timeout` instead of a single batch deadline. On any spawn/socket/restore failure, all successfully restored siblings are still dropped, so the change reports partial failures but does not preserve partial success. Orphan detection is logging-only and neither cleans up nor gates a restore that is known likely to collide. Controller restart/reconciliation remains unchanged as well.

Please use a concurrent/shared-deadline wait, define and implement an explicit partial-success contract (or narrow the PR claim), and make orphan/restart handling actionable and race-safe before treating #301 as resolved. Add tests for multiple slow/mixed-success children and restart with a real orphaned Firecracker. Thanks for the contribution—the pieces here are useful, but the current behavior still matches the failure modes the PR says it removes.

WaylandYang (changes requested):
Thank you for adding concurrent socket waits, structured partial success, and an orphan metric. These are useful improvements, but the current head still does not satisfy the PR's resilience claim or #301's acceptance criteria, and it conflicts with main with no CI.

Blocking issues:

1. Orphan handling remains observation-only. Incrementing a metric does not reap or gate on an untracked Firecracker, so a restore known to be at risk still proceeds. Rebase with the startup recovery work where applicable, implement a safe cleanup/gating path for truly untracked runtime orphans, or narrow this PR and keep #301 open.
2. `ForkResult.children` compacts successful VMs while only failures retain `child_index`; callers cannot reliably map each surviving VM back to its requested child/resource slot. Return an indexed success type (for example `{ child_index, vm }`) or an equivalent explicit mapping.
3. The controller still drops every successful sibling when any non-retryable partial failure occurs, contradicting the claim that one child no longer dooms the batch. Either expose/retain partial success through the controller API or explicitly define the controller as all-or-nothing and narrow the PR/issue-closing claim.
4. The integration test uses `n = 2` with `per_child_netns = false`, a shared-TAP/shared-IP topology that #281/#300 reject and that can fail independently with EBUSY. It also pre-creates a directory, which satisfies the current `Path::exists()` socket wait and shifts the error to restore. Use a valid per-child-netns setup or an injected spawn/socket failure and assert the exact intended phase and original child mapping.

Please resolve the main conflicts, address these contract/test issues, keep #301 open until its restart/orphan criteria are actually complete, and add the README-required DCO sign-offs to every resulting commit. Thanks for the contribution; the partial-result machinery is promising, but its public semantics need to be unambiguous and tested in a valid topology.

WaylandYang (approved):
Re-reviewed the narrowed restore-path scope at the current head. Successful children retain their original child_index, the partial-failure integration setup now uses a valid per-child-netns topology, socket waits are concurrent, all callers handle ForkChild correctly, and the controller explicitly preserves all-or-nothing semantics while keeping issue #301 open. CI is green and I found no remaining blocker. Approved. Non-blocking follow-up: classify a partial batch as retryable only when all failures are busy, so a mixed permanent + busy result does not trigger avoidable whole-batch retries.

WaylandYang:
#281 has now merged after a fresh latest-main CI run. I rechecked this branch against current main and #302 now has real content conflicts in `crates/forkd-controller/src/http.rs` and `crates/forkd-controller/src/lib.rs`.

Please rebase onto current main and preserve both sides of the controller integration:

- keep main/#281's shared-TAP claim/commit/RAII lifecycle, `SharedTapBusy` handling, and the three-part spawn result carrying `ForkResult`, `NetnsReservation`, and `SharedTapClaim`;
- keep #302's concurrent socket waits, indexed partial results, controller all-or-nothing retry/error handling, configurable timeout, and orphan-detection metric/state.

In particular, please do not resolve the tuple/AppState conflicts by dropping either the TAP ownership fields or the orphan metric. Rerun the full CI suite and retain DCO sign-offs on every contributor commit. I did not find a new blocker in the VMM portion; this is an integration rebase required by the merge order, and I will re-review promptly once the rebased head is green.

WaylandYang:

jrimmer:
Thanks for the rebase guidance — #302 is rebased onto `dev` (== main, `52f40da`) and force-pushed. The 10 commits were squashed into one (`f4119ec`) so the #281-vs-#302 conflict in `http.rs`/`lib.rs` was resolved once, cleanly, with both sides preserved as you asked.

**Both sides preserved**

#281's shared-TAP lifecycle (kept in `create_sandbox` and the workspace spawn paths):
- `SharedTapClaim` struct + `try_claim_shared_tap` + `release_shared_tap_if_owner`
- `tap_token_for_closure` clone passed into `spawn_blocking`
- the 3-part spawn result: `(ForkResult, Option<NetnsReservation>, Option<SharedTapClaim>)`
- claim returned **uncommitted**, `claim.commit()` after `live_vms.insert()` in all three spawn paths (create_sandbox, create_workspace, resume_workspace)
- `first_id_override`/`tap_token` — the first child's sandbox id is the tap-lease token
- `SharedTapBusy` → 503, n>1 shared-TAP rejection

#302's restore-path resilience (kept):
- `warn_orphan_firecrackers` pre-restore scan (point-in-time, diagnostic-only, never kills) + `orphan_firecrackers_detected` atomic counter exposed as `forkd_orphan_firecrackers_detected_total` metric
- partial-success contract: `ForkResult.children: Vec<ForkChild{vm, child_index}>` + `failures: Vec<RestoreFailure>`; `Ok(r) if r.failures.is_empty()` returns the 3-part tuple, otherwise drops the partial set and retries on busy / surfaces `RestoreError`
- concurrent socket waits + `socket_wait_timeout_secs: 30` for the daemon path
- `last_err` tracking across retry attempts

**Conflict resolution detail** (`crates/forkd-controller/src/http.rs`): the orphan scan block runs before `spawn_blocking` in both `create_sandbox` and `spawn_one_for_workspace`; the spawn tuple stays 3-part so `tap_claim` flows through to the commit-after-insert; the children loop destructures `ForkChild` (`let vm = fc.vm;`) and keeps `first_id_override.take()` for shared-tap ownership; test-state construction (`test_state_with_cap`, `test_state_with_netns`) initializes both `shared_tap_owner` and `orphan_firecrackers_detected`. `lib.rs` `AppState` gets both fields.

All commits DCO-signed. The branch is now on `dev` with a clean single-commit diff over `dev` (4 files: http.rs, lib.rs, main.rs, forkd-vmm/lib.rs). Could you re-review the rebased head? The original approve stood on the behavioral content; this rebase just integrates it with #281's merged shared-TAP stack.

WaylandYang (changes requested):
Thanks for resolving the #281 rebase and preserving both sides. The current head is not reviewable for approval yet because the required `rust` job fails under `-D warnings`: `last_err` is declared at `crates/forkd-controller/src/http.rs:1398` and assigned at line 1439, but never read. Please either remove the dead tracking or, preferably, include the final restore error in the exhausted-retry response/log so the variable serves its diagnostic purpose. Push the fix and rerun CI; I will re-review the updated head.

jrimmer:
Rebased onto current `dev` (77d3f5d) and addressed the r7 review round.

## What was addressed

**`last_err` dead code (`-D warnings` failure):** the busy-retry loop in `crates/forkd-controller/src/http.rs` now falls through on the final attempt rather than returning early, so the exhausted-retry error carries retry-exhaustion context into the API response/log. `last_err` is read at the post-loop `Err(last_err.expect(...))`.

Behavior preserved:
- Non-busy errors still return immediately (no retry, no backoff).
- Busy errors retry with the existing backoffs `[50, 200, 800]` ms.
- On exhaustion, the final busy error surfaces wrapped with `"restore_many: exhausted all busy-retries"` context — operators can now distinguish a transient contention storm from a permanent failure.
- `last_err` is always `Some` when the loop falls through (every fall-through path is a busy branch that just assigned it), so the `expect` cannot panic in normal control flow.
- No partial-success VMs are leaked on retry: `drop(children)` runs before every retry/return, and `RestoreError` holds only diagnostic `failures` (never live `Vm`s).

## Rebase

Rebased onto `dev` (77d3f5d), preserving both #281's shared-TAP lifecycle and #302's restore-path resilience:
- `sync_guest_clocks: true` (from #300 on dev) + `socket_wait_timeout_secs: daemon_socket_wait_timeout_secs()` (from this PR) on both daemon spawn paths.
- `ForkOpts`/`ForkResult` carry both sides' fields (`socket_wait_timeout_secs`, `sync_guest_clocks`, `clock_sync_ms`, `clock_sync_outcomes`, `failures`, `ForkChild`).
- `sync_guest_clocks` now takes `&[&Vm]` and is passed the flattened live children (`Vm` is not `Clone` — it owns a Child), preserving the partial-success `Vec<Option<Vm>>` shape.

All commits DCO-signed. The rebased head is `1010559`. Could you re-review?

WaylandYang (changes requested):
The rebase and busy-retry diagnostic fix are sound, and required CI is green. One partial-success correctness blocker remains in the new indexed-result contract.

After failures, `restore_many_with` builds `live` with `children.iter().flatten()` and passes that compressed slice to `sync_guest_clocks`. That function enumerates the compressed slice and writes the enumeration index into `ClockSyncOutcome`. Therefore, if slot 1 fails and slot 2 survives, the clock outcome for the real child 2 is reported as index 0/1 rather than its original within-batch slot 2. This contradicts the nearby comment that outcomes preserve the original 1-based child index, and it breaks exactly the resource/outcome correlation this PR introduces with `ForkChild` and `RestoreFailure`.

Please carry `(original_child_index, &Vm)` into clock sync (or remap outcomes without flattening away the slot), use one documented indexing convention—preferably the same 1-based convention as `ForkChild`/`RestoreFailure`—and add a regression with a hole before a surviving child. The test should prove the survivor's clock outcome retains its original slot rather than being renumbered by filtering.

Once that is fixed on current `dev` and CI is green, this looks ready for approval. The exhausted busy-retry path now correctly uses `last_err` and exposes the final context.

jrimmer:
Rebased onto current `dev` (d2d238a) and addressed the review.

## What was addressed

**Clock-outcome index renumbering (the blocker):** `restore_many_with` used to build `live` via `children.iter().flatten()`, so after a partial failure (a `None` slot) `sync_guest_clocks`'s enumeration renumbered every survivor — a child that survived in original slot 2 was reported as index 0/1, breaking the resource correlation `ForkChild`/`RestoreFailure` promise.

- `sync_guest_clocks` now takes `&[(child_index, &Vm)]`, where `child_index` is the **original 1-based within-batch slot**. It echoes that slot into `ClockSyncOutcome` and the tracing `child` spans.
- Extracted `live_children_with_indices()` — the single index-preserving mapping (`enumerate + filter_map, i+1`) used by `restore_many_with`, made unit-testable.
- Regression: `live_children_with_indices_preserves_survivor_slot_across_hole` proves `[None,Some,None,Some] -> [(2,&10),(4,&30)]`, not `[0,1]`; plus trailing-hole, first-live-then-holes, empty, all-None cases.
- Updated the single-child clock-sync integration test to pass slot 1 and assert `child_index == 1`.

One documented indexing convention now holds everywhere: `child_{i}.sock`, `ForkChild.child_index`, `RestoreFailure.child_index`, and `ClockSyncOutcome.child_index` are all 1-based within-batch.

## Rebase

Rebased onto `dev` (`d2d238a`, includes #312). Both #281's shared-TAP lifecycle and #300's clock-sync survive; `sync_guest_clocks` keeps the partial-success `Vec<Option<Vm>>` shape.

All commits DCO-signed. Head is `f1e0afa`. Could you re-review?

WaylandYang (approved):
The clock-outcome indexing blocker is properly resolved. `sync_guest_clocks` now takes `&[(child_index, &Vm)]` and echoes the original slot into both `ClockSyncOutcome.child_index` and the tracing spans, and `live_children_with_indices` isolates the index-preserving mapping (`enumerate + filter_map, i + 1`) as the single source of truth. The `live_children_with_indices_preserves_survivor_slot_across_hole` regression proves `[None, Some, None, Some] -> [(2, _), (4, _)]` without needing KVM, and the trailing-hole / first-live / empty / all-None cases cover the boundaries. `ForkResult` construction enumerates before filtering, so `ForkChild.child_index` carries the original slot too — one 1-based within-batch convention now holds across `child_{i}.sock`, `ForkChild`, `RestoreFailure`, and `ClockSyncOutcome`.

Rebased onto current `dev` (d2d238a), both #281's shared-TAP ownership lifecycle and #300's clock sync survive, and the exhausted busy-retry path correctly surfaces `last_err`. Required `rust` check and all other CI are green.

Approving and merging. Note that #301 stays open: this PR covers the per-VM partial-failure reporting and configurable timeout, while orphan reap/gating and restart recovery still depend on #299/#304 and further controller work.

PR #303, opened 2026-08-13
fix(controller): serialize registry persistence

light-merlin-dark:
## Summary

- serialize controller registry persistence across snapshot, temporary-file write, and rename
- prevent concurrent mutations from racing on the shared `state.json.tmp` path
- add a concurrent mutation regression test covering 20 workers with 10 writes each

## Root cause

The registry's in-memory map lock was released before `flush()`. Concurrent API mutations could therefore clone different generations of state and write the same `state.json.tmp` path at the same time. One caller could rename the temporary file while another was still expecting it, producing:

```text
rename .../state.json.tmp -> .../state.json: No such file or directory
```

Even when both renames succeeded, an older cloned generation could overwrite a newer one.

The new persistence mutex intentionally covers state cloning, write, and atomic rename so on-disk generations follow mutation order.

## Verification

- regression test fails against current `main` before the fix
- regression test passes with the fix
- `cargo fmt --check`
- 72 controller unit tests pass
- 7 controller HTTP integration tests pass
- live Firecracker host: spawn 20 sandboxes, issue 20 parallel DELETE requests, receive 20/20 HTTP 204 in 113 ms
- after the live run: API sandbox count 0, persisted registry sandbox count 0, DELETE 500 count 0, no Firecracker/controller process left, controller port closed

Tested from current `main` at `090ab2f` on Linux 6.8 with KVM and Firecracker 1.10.1.

By Merlin ([rbeckner.com](https://rbeckner.com)) and his AI agent (Codex Desktop, `gpt-5.6-sol`, thinking on: `xhigh`).

### DIFF
--- a/crates/forkd-controller/src/state.rs
+++ b/crates/forkd-controller/src/state.rs
@@ -1,8 +1,9 @@
 //! In-memory VM registry, snapshotted to a JSON file for crash recovery.
 //!
-//! Concurrency: a single `parking_lot::Mutex` wraps the whole registry.
-//! Writes are infrequent (one per sandbox lifecycle event) so contention
-//! is a non-issue at our scale (≤ a few thousand sandboxes/host).
+//! Concurrency: one `parking_lot::Mutex` wraps the in-memory registry and a
+//! second serializes atomic persistence. Writes are infrequent (one per
+//! sandbox lifecycle event) so contention is a non-issue at our scale
+//! (≤ a few thousand sandboxes/host).
 //!
 //! On startup, the daemon reads `state.json`, then reconciles each entry
 //! against the live system (does the netns still exist, is the FC pid
@@ -34,6 +35,7 @@ pub struct PersistentState {
 #[derive(Clone)]
 pub struct Registry {
     inner: Arc<Mutex<PersistentState>>,
+    persist_lock: Arc<Mutex<()>>,
     path: PathBuf,
 }
 
@@ -54,6 +56,7 @@ impl Registry {
         };
         Ok(Self {
             inner: Arc::new(Mutex::new(state)),
+            persist_lock: Arc::new(Mutex::new(())),
             path,
         })
     }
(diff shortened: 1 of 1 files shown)


WaylandYang (changes requested):
Thank you for the focused fix and the strong live verification. I reviewed the locking order and persistence sequence: sharing one persistence mutex across Registry clones and re-reading state inside the lock correctly prevents an older queued flush from overwriting a newer generation. I did not find a code-level correctness blocker.

Before merge, please amend the commit with the project-required DCO trailer (`git commit -s`, producing `Signed-off-by: ...`) and force-push the updated commit. The current commit has no sign-off. I have approved the first-contributor Actions run; after the signed-off commit is pushed and all CI checks pass, this should be ready to approve. A deterministic scheduling test could be a follow-up and is not blocking. Thanks for the contribution.

WaylandYang (changes requested):
The code-level fix remains sound: the persistence mutex is shared by Registry clones and the state snapshot is taken inside that lock, preventing stale queued flushes and shared-temp-file races. The current commit still has no repository-required DCO Signed-off-by trailer, and the branch is behind main. Please rebase or update the branch, amend the commit with git commit -s, and rerun CI. No additional code blocker found.

light-merlin-dark:
@WaylandYang Done — commit amended and force-pushed as `f0a13e7`:

- Rebased onto current `main` (the branch was behind); no file overlap with the intervening commits, so the patch is byte-identical to the version you reviewed (`state.rs` diff against `0534ae1` is empty).
- Amended with `git commit -s`; the commit now carries `Signed-off-by: light-merlin-dark <merlin@threewizards.agency>`.
- `cargo fmt --check` passes locally.

The new CI run is queued as `action_required` (first-time-contributor gate on the fresh push) — could you approve the Actions run when you get a chance? Thanks for the quick review.

By Merlin ([rbeckner.com](https://rbeckner.com)) and his AI agent (Kimi Code CLI, `kimi-code/k3`, thinking on).

WaylandYang (approved):
Re-reviewed f0a13e7 after the rebase. The patch is byte-identical to the code-level version already reviewed, the commit is based on current main and now carries the required DCO sign-off. The new CI run is fully green, including fmt, clippy, build, Rust tests, and all Python jobs. The shared persistence lock and in-lock state snapshot correctly close the temp-file and stale-flush race. Approved — thank you for the clean follow-up.

PR #312, opened 2026-08-22
fix(controller): run Registry::reconcile under a single lock

### LINKED ISSUE #310: `Registry::reconcile` re-acquires the lock for every stale entry
In `crates/forkd-controller/src/state.rs:210-256`, the reconcile pass does roughly this:

1. lock; build `stale: Vec<String>` of sandbox ids whose pid is dead.
2. for each id in `stale`: lock; remove.
3. lock; collect live ids.
4. lock; walk workspaces, mark stale ones.
5. flush (clones state under a fifth lock).

So a reconcile of `K` stale sandboxes takes `K + 4` lock acquisitions. Worse, between steps 2 and 3 nothing prevents an HTTP handler from inserting a new sandbox whose pid is alive; that's harmless. But between step 1 and step 2, an HTTP handler could `insert_sandbox` reusing an id (unlikely with the monotonic counter, but possible) or `remove_sandbox` it first — which would mean step 2's removal returns `None` and we incorrectly account for it.

This is a low-traffic path so it's never going to dominate performance, but the lock-thrashing pattern is the kind of thing that bites you later when reconcile is called from a periodic task that fires alongside spawn bursts.

**Suggested shape:**

```rust
pub fn reconcile(&self) -> Result<usize> {
    let (pruned, ws_changed) = {
        let mut g = self.inner.lock();
        // do everything inside one lock; pid_alive is a /proc stat
        // which is cheap and doesn't hold any kernel locks we care about
        let stale_ids: Vec<_> = g.sandboxes.iter()
            .filter_map(|(id, sb)| match sb.pid {
                Some(pid) if !pid_alive(pid) => Some(id.clone()),
                _ => None,
            })
            .collect();
        for id in &stale_ids { g.sandboxes.remove(id); }
        let live_ids: HashSet<_> = g.sandboxes.keys().cloned().collect();
        let mut ws_changed = false;
        for ws in g.workspaces.values_mut() {
            if ws.status == WorkspaceStatus::Running {
                let live = ws.live_sandbox_id.as_ref()
                    .is_some_and(|id| live_ids.contains(id));
                if !live {
                    ws.status = WorkspaceStatus::Stale;
                    ws.live_sandbox_id = None;
                    ws_changed = true;
                }
            }
        }
        (stale_ids.len(), ws_changed)
    };
    if pruned > 0 || ws_changed { self.flush()?; }
    Ok(pruned)
}
```

`pid_alive` is doing a stat on `/proc/{pid}` — fast, doesn't take any contended locks; safe to call under the registry mutex. The single-lock shape removes the lost-update risk and shrinks the function noticeably.

AmirF194:
## Summary

`Registry::reconcile()` took `K + 4` separate registry-mutex acquisitions to prune `K` stale sandboxes: one lock per removed id, plus one each for the initial stale scan, the live-id snapshot, and the workspace walk. As you noted in the issue, nothing stops an HTTP handler from inserting or removing a sandbox between the scan and the per-id removal, which can make the reported prune count diverge from what was actually removed.

This runs the whole in-memory pass (stale scan, removal, live-id snapshot, workspace walk) under a single `self.inner.lock()` acquisition, matching the shape from the issue. `pid_alive` is a `/proc` stat and takes no lock the registry cares about, so holding the mutex across it is fine. `flush()` behavior is unchanged: still called once, after the lock is dropped, only when `pruned > 0 || stale_ws_changed`.

## Why not a one-liner

The function reads sandboxes then mutates workspaces from the same `MutexGuard`; the live-id set has to stay an owned `HashSet<String>` (not borrowed) so the borrow checker allows the later mutable walk over `g.workspaces` through the same guard.

## Verification

- New test `reconcile_prunes_dead_sandbox_and_marks_its_workspace_stale`: a sandbox with a dead pid is pruned and its `Running` workspace flips to `Stale` with `live_sandbox_id` cleared; a sandbox with no recorded pid (`pid: None`) is left alone and its workspace stays `Running`; a `Suspended` workspace is never touched even though its sandbox is gone; reloading from disk confirms `flush()` ran.
- New test `reconcile_no_op_does_not_flush`: a reconcile pass that prunes nothing leaves `state.json` byte-identical.
- All four `rust` CI steps run clean in a `rust:1-bookworm` container against this branch (94 passed, 0 failed, in the `forkd-controller` unit crate):
  ```
  cargo fmt --all -- --check
  cargo clippy --all-targets --all-features -- -D warnings
  cargo build --all
  cargo test --all
  ```
- Not verified: the actual interleaving/lost-accounting scenario from the issue. As you said, `reconcile()` only runs at startup today before the HTTP listener accepts requests, so it is not reachable through any current call path; the two new tests cover the prune-count and workspace-transition behavior the single-lock refactor must preserve, not a live race.

## Note on #299

#299 is open on the same function (extracts the workspace-marking half into `mark_stale_workspaces()`); it does not touch the stale-id removal loop this issue is about, so it is not a competing fix, but merging either PR first will need a small rebase of the other.

Fixes #310

### DIFF
--- a/crates/forkd-controller/src/state.rs
+++ b/crates/forkd-controller/src/state.rs
@@ -216,35 +216,34 @@ impl Registry {
         Ok(())
     }
 
-    /// Prune sandbox entries whose recorded pid is no longer alive.
-    /// Snapshots are kept regardless (they're disk artifacts).
+    /// Prune sandbox entries whose recorded pid is no longer alive, and
+    /// mark any `Running` workspace whose sandbox was just pruned as
+    /// `Stale` (#116): the daemon crashed/restarted out from under it.
+    /// Suspended workspaces are untouched; they were intentionally
+    /// parked. Snapshots are kept regardless (they're disk artifacts).
+    ///
+    /// Runs the whole in-memory pass under one registry lock (#310):
+    /// `pid_alive` is a `/proc` stat, cheap and free of any lock we
+    /// care about, so holding the mutex across it is fine and removes
+    /// the window where an HTTP handler could remove/reinsert a
+    /// just-observed-stale id between the scan and the prune.
     pub fn reconcile(&self) -> Result<usize> {
-        let mut pruned = 0usize;
-        let stale: Vec<String> = {
-            let g = self.inner.lock();
-            g.sandboxes
+        let (pruned, stale_ws_changed) = {
+            let mut g = self.inner.lock();
+            let stale_ids: Vec<String> = g
+                .sandboxes
                 .iter()
                 .filter_map(|(id, sb)| match sb.pid {
                     Some(pid) if !pid_alive(pid) => Some(id.clone()),
                     _ => None,
                 })
-                .collect()
-        };
-        for id in stale {
-            self.inner.lock().sandboxes.remove(&id);
-            pruned += 1;
-        }
+                .collect();
+            for id in &stale_ids {
+                g.sandboxes.remove(id);
+            }
 
-        // Workspaces (#116): any workspace marked Running whose
-        // live_sandbox_id is no longer in the live sandbox table is
-        // Stale — the daemon crashed/restarted out from under it.
-        // We don't touch Suspended workspaces; they were intentionally
-        // parked.
-        let live_ids: std::collections::HashSet<String> =
-            self.inner.lock().sandboxes.keys().cloned().collect();
-        let mut stale_ws_changed = false;
-        {
-            let mut g = self.inner.lock();
+            let live_ids: std::collections::HashSet<String> = g.sandboxes.keys().cloned().collect();
+            let mut stale_ws_changed = false;
             for ws in g.workspaces.values_mut() {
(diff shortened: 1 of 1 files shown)


WaylandYang (approved):
The single-lock refactor matches the accepted shape from #310: stale discovery/removal, the live-id snapshot, and Running→Stale workspace transitions now occur under one registry guard, while persistence remains outside the lock and conditional on an actual change. The focused tests cover prune accounting, workspace transitions, suspended/no-PID preservation, and reload persistence. DCO and all required CI checks are green.

Approved. Non-blocking: the byte-equality no-op test confirms stable output but cannot by itself prove no rewrite occurred; the explicit `if pruned > 0 || stale_ws_changed` control flow is clear enough here.

AmirF194:
Thanks for the quick review and merge. Glad the single-lock shape matched what you had in mind for #310.

WaylandYang:
Thanks for the contribution, Amir — welcome aboard!

Issue #317, opened 2026-09-11
Restored children share one writable rootfs — concurrent sandboxes corrupt each other


jrimmer:
Restores are not safe to run concurrently against one snapshot, and the failure is silent corruption rather than an error. Filing with the evidence because it cost us ~12h to attribute, and because the fix is a design decision I would rather ask you about than assume.

## Mechanism

1. An rw bake boots the parent with `rootfs_read_only: false` (`BootConfig::ext4_rw`, `crates/forkd-vmm/src/lib.rs:159`; the ro variant `ext4_ro` at `:139` sets `true`).
2. Firecracker serialises the drive's `path_on_host` + `is_read_only` into the vmstate, and reopens that path verbatim on `/snapshot/load`.
3. `snapshot.json` pins the rootfs as an absolute **shared** path, not per-snapshot or per-child:

```json
{ "vmstate": ".../elixir-release/vmstate",
  "memory":  ".../elixir-release/memory.bin",
  "rootfs":  "/var/cache/forkd/elixir-release-tools-local.ext4" }
```

4. `restore_many_with` copies/layers **memory only** — the v0.5 chain assembly stages `memory-assembled.bin`; the rootfs path is untouched.

So every restored child of a tag opens the same ext4 read-write. Two children, two guest kernels, one filesystem, no coordinator:

```
$ lsof | grep firecrack
firecrack 2588644 root 30u REG 252,1 12884901888 4628036 /var/cache/forkd/elixir-release-tools-local.ext4
firecrack 2600514 root 30u REG 252,1 12884901888 4628036 /var/cache/forkd/elixir-release-tools-local.ext4
```

Same inode, both `30u` (read-write).

## Observable damage

Random, different victims each run, and the base ext4 still hashes byte-identical to its source image in the same session — so nothing points at the image:

- `/usr/bin/uname` holding another file's bytes — a valid ELF of plausible size, so an "it still executes" check passes; only an identity check catches it (`uname --version` → `uniq (GNU coreutils) 9.1`). `/usr/bin/tr` likewise.
- `libapt-pkg.so.6.0: invalid ELF header`, `libidn2.so.0: cannot dynamically load position-independent executable`.
- `EBADMSG` on `/var/lib/apt/lists/*` and `/var/lib/dpkg/triggers/File` — corrupt directory entries.
- The ext4's mtime changing while no bake is running.

Two consequences worth separating:

[...]

PR #318, opened 2026-09-11
fix(controller): detect and warn on a shared writable rootfs

jrimmer:
Detection and an opt-in guard for the defect in #317 — the bridge, not the per-child writable layer, because the layer turns out to hinge on a question I would rather you answer than assume.

## What I found that changes the options in #317

I proposed a host-side per-child reflink there. **That option does not exist**: Firecracker serialises the drive's `path_on_host` *and* `is_read_only` into the vmstate and reopens both verbatim at restore, and there is no way to override either at load time. I searched upstream v1.12 for `drive_overrides` and there is no such field, so the host cannot re-point a child's rootfs, and forkd's JSON path rewriting cannot reach it because the vmstate is binary. Host-side per-child copies would need a Firecracker feature first.

That leaves the guest side, and the guest side has the same shape of problem: the boot path cannot tell a child from a parent either — the kernel command line is frozen in the vmstate too. So a per-child layer needs a mechanism for a restored guest to learn it is a child (and then mount the shared base read-only with its own upper). That is a design decision, not an implementation detail, which is why #317 asks rather than assumes.

## What lands here

The part every option needs, and that is safe on its own.

- `Snapshot.rootfs_read_only: Option<bool>` — purely additive, `serde` defaulted and skipped when absent, so existing `snapshot.json` files stay byte-identical (the same pattern as `volumes` / `parent_tag` / `rootfs`). The CLI bake records the flag it actually booted with.
- `Snapshot::rootfs_is_read_only()` — the recorded flag wins; for pre-flag snapshots an `.ext4` rootfs is *inferred* writable, because that is exactly the convention the boot path already uses to choose `BootConfig::ext4_rw` (`rw_flag || extension == "ext4"`). Anything else returns `None`, which callers must read as "cannot vouch for it" rather than "safe".
- The controller warns on any restore that would add a second writer to a writable-rootfs snapshot, naming the live holders and both remedies (re-bake read-only, or one child at a time). `FORKD_REFUSE_SHARED_RW=1` upgrades that to a `409`.

## Why the refusal is opt-in

Refusing rejects `fork -n N` on a writable snapshot — and that is the flow `from-image` prints as its next step, with `rw` hardcoded, while `restore_many_with` is built to share the inode across n children. Making that a hard error by default is a behaviour change that belongs to you, not to me. Set the flag and the current "one writer at a time" discipline becomes an enforced invariant; leave it unset and operators get the signal without a broken workflow.

## Verification

`cargo fmt --check` and `cargo clippy --all-targets --all-features -D warnings` clean; `cargo test -p forkd-vmm --lib` 55 passed, `-p forkd-controller --lib` 114 passed. A unit test covers the flag resolution (recorded wins both ways; `.ext4` inferred; non-ext4 and no-rootfs stay unknown).

I have not written a test that exercises the warning through `create_sandbox` — building an `AppState` for it is heavy, and I would rather add it in the shape you would want than guess. Deployed on our own host with `FORKD_REFUSE_SHARED_RW=1` as the interim.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,33 @@ snapshot is fully published, so a tag always holds a consistent
 src == dst guard comment duplication and makes the chain-unpack tail
 bail instead of resolving an empty path.
 
+### Snapshots record the rootfs drive's read-only flag
+
+`snapshot.json` now carries `rootfs_read_only`, the flag in effect when
+the vmstate was frozen. It matters because Firecracker serialises the
+drive's `path_on_host` and `is_read_only` into the vmstate and reopens
+both verbatim at restore, with no way to override either at load time —
+the API has no `drive_overrides` and the vmstate is a binary blob, not
+JSON. A snapshot baked with a writable rootfs therefore hands *every*
+restored child the same ext4 opened read-write, and two concurrent
+children are two guest kernels writing one filesystem with no
+coordinator: package files pick up other files' bytes, directory entries
+go `EBADMSG`, and the damage surfaces as a random build failure rather
+than a sandbox error.
+
+The controller detects this and warns when a restore would add a second
+writer to a writable-rootfs snapshot, naming the live holders and the
+remedies (re-bake read-only, or restore one child at a time). With
+`FORKD_REFUSE_SHARED_RW=1` the warning becomes a `409`. Refusing is
+opt-in because it rejects `fork -n N` on a writable snapshot — the flow
+`from-image` prints as its next step — and that behaviour change should
+be the operator's choice.
+
+Snapshots written before this field existed carry no flag;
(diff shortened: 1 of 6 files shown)


jrimmer:
Closing this: #319 supersedes it, and keeping it would be actively wrong.

This warned — or refused, under `FORKD_REFUSE_SHARED_RW=1` — when a restore would add a second writer to a writable-rootfs snapshot. With #319 the guest never writes the shared base at all (`/` is mounted `ro`, writes go to a tmpfs/overlay layer in guest RAM), so concurrency is safe and the refusal would be a false positive blocking exactly the parallelism the fix restores.

The `rootfs_read_only` field it added to `snapshot.json` is therefore no longer needed for safety either; I dropped it from #319 rather than carry an unused field. The branch stays on my fork if you ever want the drive flag recorded for diagnostics.

